PR #33, opened 2018-10-07
More idiomatic value assignment

luisbg:
If the value of a variable is based on a condition it is more idiomatic
to use let x = if { a } else { b }; Plus more readable too.


brycx:
Thank you for this PR @luisbg. It looks very good and also brought two other points of improvement to my attention:
1. We should be able to remove the `.to_le()` call on `x`, in `left_encode()`
2. The documentation is inconsistent with the rest of the library

If you want to fix the first, you can just add it to this PR. I've created an issue [here](https://github.com/brycx/orion/issues/34) in case you'd want to take a shot at the second one as well.

luisbg:
I will tackle number 1 and add it to this PR. Thanks for reviewing :)

brycx:
I'm merging this and adding improvement nr. 1 as well.

PR #45, opened 2018-12-14
Use subtle v2

hdevalence:
As far as I can tell, the `ConstantTimeEq` trait is only used internally, and is not implemented on any public types, so this should be a semver-compatible change that does not require incrementing the major version.


brycx:
I see no problem with this either. Travis CI fails, but when building locally with `rustc 1.32.0-nightly` instead of `1.33`, no problems are reported with target `thumbv7em-none-eabihf`.

Thank you for the PR @hdevalence, and thank you for the great work on the dalek crates!

Issue #77, opened 2019-07-13
Add usage example for newtypes and their PartialEq implementations


brycx:
Documentation should show how users can safely compare slices and newtypes with each other. The documentation should also include a short note on security, where constant-time comparison is required.

brycx:
The only examples now missing after #78 (documented `newtype == &[u8]`) are examples for `newtype == newtype`.

PR #78, opened 2019-07-19
Document newtype PartialEq usage 

vlmutolo:
One unresolved issue is having independent documentation generated for each type that a single macro creates. I'm not currently sure how to solve this.

### DIFF
--- a/src/typedefs.rs
+++ b/src/typedefs.rs
@@ -622,7 +622,7 @@ macro_rules! construct_hmac_key {
         /// let secret_key = SecretKey::generate();
         ///
         /// // Secure, costant-time comparison with byte slice
-        /// assert!(secret_key != key_bytes);
+        /// assert!(secret_key != &[0; 32][..]);
         /// ```
         pub struct $name {
             value: [u8; $size],
--- a/src/typedefs.rs
+++ b/src/typedefs.rs
@@ -405,12 +405,10 @@ macro_rules! construct_secret_key {
         /// use orion::hazardous::stream::chacha20::SecretKey;
         ///
         /// // Initialize a secret key with 32 random bytes.
-        /// let key_bytes: [u8; 32] =
-        ///     [0x2f, 0xbd, 0x03, 0xeb, 0x36, 0x66, 0xa9, 0x8c,
-        ///      0x39, 0x7f, 0xe7, 0xc3, 0x53, 0x4b, 0x9f, 0x8e,
-        ///      0x17, 0x76, 0x5d, 0x96, 0xf4, 0x08, 0x6c, 0x8f,
-        ///      0x25, 0x98, 0x80, 0xd9, 0xb2, 0xa5, 0x56, 0xce];
-        /// let secret_key = SecretKey::from_slice(&key_bytes).unwrap();
+        /// let secret_key = SecretKey::generate();
+        ///
+        /// // Get the byte slice out of the key to demonstrate the comparison.
+        /// let key_bytes = secret_key.unprotected_as_bytes();
         ///
         /// // Variable-time comparison: INSECURE!! DO NOT DO THIS!!
         /// assert!(secret_key.unprotected_as_bytes() == key_bytes);
@@ -565,23 +563,11 @@ macro_rules! construct_tag {
         /// ```rust
         /// use orion::hazardous::mac::hmac::Tag;
         ///
-        /// // Initialize a tag with 64 random bytes.
-        /// let tag_bytes: [u8; 64] =
-        ///     [0x15, 0xf2, 0x52, 0x22, 0x01, 0x4c, 0x01, 0x85,
-        ///      0x6d, 0xe3, 0xb4, 0x94, 0x09, 0x9a, 0xe8, 0x28,
-        ///      0xf1, 0x61, 0xc8, 0xbc, 0x57, 0x2e, 0xa3, 0x1a,
-        ///      0x49, 0x1f, 0x79, 0x84, 0x93, 0x49, 0x68, 0xab,
-        ///      0xb2, 0xc8, 0xf1, 0xa7, 0xa6, 0xbc, 0x4d, 0x60,
-        ///      0x09, 0x91, 0x50, 0x22, 0xd5, 0xf9, 0xa1, 0x80,
-        ///      0x46, 0x25, 0x3c, 0x2e, 0x76, 0x79, 0xaa, 0xa1,
-        ///      0x84, 0x95, 0x70, 0x20, 0xf2, 0xfd, 0xe5, 0x16];
-        /// let tag = Tag::from_slice(&tag_bytes).unwrap();
+        /// // Initialize an arbitrary, 64-byte tag.
+        /// let tag = Tag::from_slice(&[1; 64]).unwrap();
         ///
-        /// // Variable-time comparison: INSECURE!! DO NOT DO THIS!!
-        /// assert!(tag.unprotected_as_bytes() == &tag_bytes[..]);
-        ///
-        /// // Constant-time comparison: SECURE, PREFER THIS WHEN POSSIBLE
-        /// assert!(tag == &tag_bytes[..]);
+        /// // Constant-time comparison
+        /// assert!(tag == &[1; 64][..]);
         /// ```
         pub struct $name {
             value: [u8; $upper_bound],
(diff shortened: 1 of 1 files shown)


brycx:
Closes #77.

src/typedefs.rs
```diff
@@ -395,6 +395,30 @@ macro_rules! construct_secret_key {
         /// # Security:
         /// - __**Avoid using**__ `unprotected_as_bytes()` whenever possible, as it breaks all protections
         /// that the type implements.
+        ///
+        /// - The trait `PartialEq<&'_ [u8]>` is implemented for this type so that users are not tempted
+        /// to call `unprotected_as_bytes` to compare this sensitive value to a byte string. The trait
```
brycx: `byte slice` seems more correct than `byte string`.

src/typedefs.rs
```diff
@@ -395,6 +395,30 @@ macro_rules! construct_secret_key {
...
+        /// is implemented in such a way that the comparison happens in constant time. Thus, users should
+        /// prefer `SecretType == &[u8]` over `SecretType.unprotected_as_bytes() == &[u8]`.
+        /// Examples are shown below. Although the `orion::hazardous::stream::chacha20::SecretKey`
+        /// type is used, this example applies to any fixed-size store of sensitive data.
```
brycx: It also applies to the variable-sized (heap) secret values, not only fixed-size. Generally I think this is clear from any type that implements said trait, so maybe we can get rid of "Although the [..] store of sensitive data." altogether to also keep the bullet-point a bit more concise?

brycx: Or say "The examples apply to any type that implements `PartialEq<&'_ [u8]>`. That would be more clear.

src/typedefs.rs
```diff
@@ -395,6 +395,30 @@ macro_rules! construct_secret_key {
...
+        /// ```rust
+        /// use orion::hazardous::stream::chacha20::SecretKey;
+        ///
+        /// // Initialize a secret key with 32 random bytes.
+        /// let key_bytes: [u8; 32] =
+        ///     [0x2f, 0xbd, 0x03, 0xeb, 0x36, 0x66, 0xa9, 0x8c,
+        ///      0x39, 0x7f, 0xe7, 0xc3, 0x53, 0x4b, 0x9f, 0x8e,
+        ///      0x17, 0x76, 0x5d, 0x96, 0xf4, 0x08, 0x6c, 0x8f,
+        ///      0x25, 0x98, 0x80, 0xd9, 0xb2, 0xa5, 0x56, 0xce];
+        /// let secret_key = SecretKey::from_slice(&key_bytes).unwrap();
```
brycx: We want to encourage users not to store secrets in source and use strong keys. Please remove `key_bytes` and initialize `secret_key` with `let secret_key = SecretKey::generate();`.

src/typedefs.rs
```diff
@@ -395,6 +395,30 @@ macro_rules! construct_secret_key {
...
+        ///
+        /// // Variable-time comparison: INSECURE!! DO NOT DO THIS!!
```
brycx: Including this seems a bit redundant as we already mention this is the security section. The more clean and easy-to-read we can keep the examples - the better. Let's remove this for now.

src/typedefs.rs
vlmutolo: I like the phrasing in your second example. "The examples apply to any type that implements `PartialEq<&'_ [u8]>`". Will make the change.

src/typedefs.rs
brycx: For the later `assert` you can use a `"Not random secret value".as_bytes()` and negate to `!=`.

src/typedefs.rs
```diff
@@ -395,6 +395,30 @@ macro_rules! construct_secret_key {
...
+        /// assert!(secret_key.unprotected_as_bytes() == key_bytes);
+        ///
+        /// // Constant-time comparison: SECURE, PREFER THIS WHEN POSSIBLE
```
brycx: Please don't capitalize and instead just write something like "Secure costant-time comparison with byte slice:".

src/typedefs.rs
```diff
@@ -533,6 +557,33 @@ macro_rules! construct_tag {
...
+        /// - The trait `PartialEq<&'_ [u8]>` is implemented for this type so that users are not tempted
+        /// to call `unprotected_as_bytes` to compare this sensitive value to a byte string. The trait
+        /// is implemented in such a way that the comparison happens in constant time. Thus, users should
+        /// prefer `SecretType == &[u8]` over `SecretType.unprotected_as_bytes() == &[u8]`.
+        /// Examples are shown below.
+        /// ```rust
+        /// use orion::hazardous::mac::hmac::Tag;
+        ///
+        /// // Initialize a tag with 64 random bytes.
+        /// let tag_bytes: [u8; 64] =
```
brycx: To keep the examples short and concise please remove `tag_bytes` and initialize the secret key with something like `&[0u8; 32]` and compare with the same in the `assert`.

src/typedefs.rs
```diff
@@ -533,6 +557,33 @@ macro_rules! construct_tag {
...
+        ///      0x6d, 0xe3, 0xb4, 0x94, 0x09, 0x9a, 0xe8, 0x28,
+        ///      0xf1, 0x61, 0xc8, 0xbc, 0x57, 0x2e, 0xa3, 0x1a,
+        ///      0x49, 0x1f, 0x79, 0x84, 0x93, 0x49, 0x68, 0xab,
+        ///      0xb2, 0xc8, 0xf1, 0xa7, 0xa6, 0xbc, 0x4d, 0x60,
+        ///      0x09, 0x91, 0x50, 0x22, 0xd5, 0xf9, 0xa1, 0x80,
+        ///      0x46, 0x25, 0x3c, 0x2e, 0x76, 0x79, 0xaa, 0xa1,
+        ///      0x84, 0x95, 0x70, 0x20, 0xf2, 0xfd, 0xe5, 0x16];
+        /// let tag = Tag::from_slice(&tag_bytes).unwrap();
+        ///
+        /// // Variable-time comparison: INSECURE!! DO NOT DO THIS!!
```
brycx: See above comments about variable-time comparison example and capitalisation.

src/typedefs.rs
```diff
@@ -579,6 +630,34 @@ macro_rules! construct_hmac_key {
...
+        /// - The trait `PartialEq<&'_ [u8]>` is implemented for this type so that users are not tempted
+        /// to call `unprotected_as_bytes` to compare this sensitive value to a byte string. The trait
+        /// is implemented in such a way that the comparison happens in constant time. Thus, users should
+        /// prefer `SecretType == &[u8]` over `SecretType.unprotected_as_bytes() == &[u8]`.
+        /// Examples are shown below.
+        /// ```rust
+        /// use orion::hazardous::mac::hmac::SecretKey;
+        ///
+        /// // Initialize a secret key with 32 random bytes.
+        /// let key_bytes: [u8; 32] =
```
brycx: Please see above comments about randomly generating keys. Again, here you can use `&[0u8; 128]` with a negated `assert` for comparison. This will also remove the long comment and use of `unprotected_as_bytes()`.

Please also see above comments about variable-time example and capitalisation.

src/typedefs.rs
```diff
@@ -677,6 +756,30 @@ macro_rules! construct_secret_key_variable_size {
...
+        ///
+        /// - The trait `PartialEq<&'_ [u8]>` is implemented for this type so that users are not tempted
+        /// to call `unprotected_as_bytes` to compare this sensitive value to a byte string. The trait
+        /// is implemented in such a way that the comparison happens in constant time. Thus, users should
+        /// prefer `SecretType == &[u8]` over `SecretType.unprotected_as_bytes() == &[u8]`.
+        /// Examples are shown below. Although the `orion::pwhash::Password` type is used, this example
+        /// applies to any heap-allocated store of sensitive data.
+        /// ```rust
+        /// use orion::pwhash::Password;
+        ///
```
brycx: Above comments on the other newtype macros also apply here.

brycx:
> One unresolved issue is having independent documentation generated for each type that a single macro creates. I'm not currently sure how to solve this.

AFAIK this is not solvable, expect for making another security section when calling the macro. The general approach has been to keep security information generic over where the newtypes are used. I currently can't see where it would need to be specific.

src/typedefs.rs
vlmutolo: Do you want to include a negative example? One that demonstrates the comparison could return false? Using something like `"Not random secret value".as_bytes()`, I mean.

vlmutolo: Actually, I see what you meant now. I like that idea. Making the change.

src/typedefs.rs
```diff
@@ -677,6 +722,21 @@ macro_rules! construct_secret_key_variable_size {
...
+        /// - The trait `PartialEq<&'_ [u8]>` is implemented for this type so that users are not tempted
+        /// to call `unprotected_as_bytes` to compare this sensitive value to a byte slice. The trait
+        /// is implemented in such a way that the comparison happens in constant time. Thus, users should
+        /// prefer `SecretType == &[u8]` over `SecretType.unprotected_as_bytes() == &[u8]`.
+        /// Examples are shown below. The examples apply to any type that implements `PartialEq<&'_ [u8]>`.
+        /// ```rust
+        /// use orion::pwhash::Password;
+        ///
+        /// // Initialize a password with 32 random bytes.
+        /// let password = Password::generate(32).unwrap();
```
brycx: We want to avoid examples using `.unwrap()` to comply with the [Rust-API guidelines](https://rust-lang-nursery.github.io/api-guidelines/checklist.html). Let the code example end similar to [this one](https://github.com/brycx/orion/blob/8de3292ca6523d6be7e59642ead86a61e05a6958/src/aead.rs#L75) and the `Result` can be propagated with `?`.

src/typedefs.rs
```diff
@@ -533,6 +548,21 @@ macro_rules! construct_tag {
...
+        /// - The trait `PartialEq<&'_ [u8]>` is implemented for this type so that users are not tempted
+        /// to call `unprotected_as_bytes` to compare this sensitive value to a byte slice. The trait
+        /// is implemented in such a way that the comparison happens in constant time. Thus, users should
+        /// prefer `SecretType == &[u8]` over `SecretType.unprotected_as_bytes() == &[u8]`.
+        /// Examples are shown below. The examples apply to any type that implements `PartialEq<&'_ [u8]>`.
+        /// ```rust
+        /// use orion::hazardous::mac::hmac::Tag;
+        ///
+        /// // Initialize an arbitrary, 64-byte tag.
+        /// let tag = Tag::from_slice(&[1; 64]).unwrap();
```
brycx: We want to avoid examples using `.unwrap()` to comply with the [Rust-API guidelines](https://rust-lang-nursery.github.io/api-guidelines/checklist.html). Let the code example end similar to [this one](https://github.com/brycx/orion/blob/8de3292ca6523d6be7e59642ead86a61e05a6958/src/aead.rs#L75) and the `Result` can be propagated with `?`.

src/typedefs.rs
```diff
@@ -533,6 +548,21 @@ macro_rules! construct_tag {
...
+        ///
+        /// // Constant-time comparison
```
brycx: Please make this comment consistent with the others, i.e change to "Secure, costant-time comparison with byte slice".

brycx:
CI failure seems unrelated, all tests pass for me when run locally. I made a spelling-mistake when I asked for an updated code-comment in the example, which I will fix shortly after merging this.

Thank you very much for the PR.

PR #86, opened 2019-08-31
Add CONTRIBUTING guidelines and CHANGELOG

brycx:
Adds contributing guidelines and makes the changelog more portable than it is using the GitHub API.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,71 @@
+
+### 1. What to contribute
+There are a variety of things that one may contribute to this project. Contributions of any kind are always welcomed.
+
+#### 1.1 Features and Improvements
+Improvements to the library can come in many flavors: Performance, security, usability, etc.
+An improvement that doesn't add new features or are breaking changes, may be submitted as pull requests directly.
+
+Features that add new functionality or are breaking changes should preferably be discussed in a separate issue first.
+
+Try to make changes adhere to the [Rust API Guidelines](https://rust-lang-nursery.github.io/api-guidelines/checklist.html) as much as possible.
+
+#### 1.2 Testing
+One way to contribute to testing is by writing new unit-tests to cover code that isn't already being tested. Improvements to existing unit-tests is also an option.
+
+Adding test vectors (located in `/tests`) is also a good way to improve the testing of the library. Test vectors can be official or generated using another crypto library.
+
+#### 1.3 Fuzzing
+Fuzzing is an important part of testing this library. Contributions to this aspect can come in two ways: 1) Running the fuzzing targets, updating the corpus and reporting any issues found and 2) Overall improvements to the fuzzing targets.
+
+Please refer to the [orion-fuzz](https://github.com/brycx/orion-fuzz) repository when working with fuzzing.
+
+#### 1.4 Documentation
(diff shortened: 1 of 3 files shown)


brycx:
What do you think @vlmutolo?

CONTRIBUTING.md
```diff
@@ -0,0 +1,71 @@
...
+Adding test vectors (located in `/tests`) is also a good way to improve the testing of the library. Test vectors can be official or generated using another crypto library.
+
+#### 1.3 Fuzzing
+Fuzzing is an important part of testing this library. Contributions to this aspect can come in two ways: 1) Running the fuzzing targets, updating the corpus and reporting any issues found and 2) Overall improvements to the fuzzing targets.
+
+Please refer to the [orion-fuzz](https://github.com/brycx/orion-fuzz) repository when working with fuzzing.
+
+#### 1.4 Documentation
+Quality of documentation is a vital part of this project. Contributions to this could include adding documentation where such is missing, clarifying documentation that is unclear or improving examples.
+
```
vlmutolo: You may want to mention here that it would be better to use project links instead of web links in the docs. For example, in the docs for the `auth` module, instead of writing `https://docs.rs/orion/0.14.4/orion/auth/struct.SecretKey.html`, it's better to write `struct.SecretKey.html`. This way, local docs link to local docs, and online docs link to online docs.

I'm sure there's a way to do absolute references that also don't link to online docs, but I haven't figured it out yet.

CONTRIBUTING.md
```diff
@@ -0,0 +1,71 @@
...
+  - `cargo +nightly test --tests --no-default-features --features no_std`
+
+- [ ] The formatting is correct and clippy does not show warnings by running:
+
+  - `cargo clippy`
+  - `cargo fmt`
+
+- [ ] Explain what the pull request changes, in the description of the GitHub PR.
+- [ ] If the pull request is a bugfix, try to include a regression test for the bug.
+
```
vlmutolo: Maybe it would be worthwhile to include that people should make sure to put back the clippy deny compile macro if they removed it to back their project compile before it was finished.

CONTRIBUTING.md
brycx: Sounds good to me. Actually, I think it would be good to adopt this approach throughout all of orion. Do you want to open an issue for this, and perhaps give some examples on how something would be changed to the new format?

CONTRIBUTING.md
brycx: Good idea.

CONTRIBUTING.md
brycx: Seems like the all the relevant information for this is [here](https://rust-lang-nursery.github.io/api-guidelines/documentation.html#prose-contains-hyperlinks-to-relevant-things-c-link) and [here](https://github.com/rust-lang/rfcs/blob/master/text/1574-more-api-documentation-conventions.md#link-all-the-things).

brycx: Because section 1.4 already mentions the Rust API guidelines and these explain this linking, I don't think it's necessary to make another point specifically about this.

vlmutolo: That’s true about the Rust guidelines. Sounds good. If you haven’t yet, I’ll open an issue and get to work on the PR. I have a feeling it won’t be much work.

PR #87, opened 2019-08-31
More explicit struct initialization

brycx:
Currently, for a given streaming context `Ctx` in module `foo` we have:
```
let _: Ctx = foo::init();
```
This PR changes this to be more explicit:

```
let _: Ctx = foo::Ctx::init();
```


brycx:
What do you think @vlmutolo?

vlmutolo:
I think that the explicit version is much better. It definitely helps to immediately what the type of the newly-created object is.

The only question I have is whether it would be better to rename `init` to `new`, as is the Rust convention. But that would only be if the structs aren't ever really created in any other way. Meaning, is `init` the standard way to create these objects? I think it is, and therefor should be called `new`. But maybe I'm missing something.

brycx:
> The only question I have is whether it would be better to rename init to new, as is the Rust convention.

I agree this seems to be the way in Rust, though from what I have seen, using `init()` is more common throughout other crypto libraries which are not in Rust. Maybe the Rust API guidelines have something on this?

vlmutolo:
Hey, quick turnaround! Nice. I think that this will be more intuitive for new users of the library. Less cognitive load wondering if `init()` is something special.

brycx:
Unfortunately it seems to be a little more trouble than this. I didn't realize it, but clippy warns against `new()` constructors that don't return a `Self` or use `&self`. So `Hasher` in `blake2b` won't be able to have a `new()` method, but keep the old `init()`. We'll also need to add `impl Default` to `Sha512`. Adding `Default` is fine, but having `Hasher` being inconsistent with the rest seems a bit weird no?

Here's the CI output: https://travis-ci.org/brycx/orion/jobs/580625988

brycx:
Taking into account that the `Hasher` in fact initializes a new `Blake2b` struct, I think it might not be that unclear after all. I'll accommodate the clippy suggestions, and keep the struct initialization to `new()`.

Issue #58, opened 2019-02-07
Return-type of BLAKE2b in keyed and non-keyed mode is the same


brycx:
Currently BLAKE2b always returns a `Digest`, even if it was used with a secret-key and represents a MAC. `Digest` does implement `PartialEq` in constant-time, like `Tag` in HMAC and Poly1305, but doesn't provide the function `unprotected_as_bytes()`, but instead `as_bytes()`.

Perhaps BLAKE2b should return a `Tag` when used with a secret key, making the difference more clear. Maybe even move keyed-BALKE2b into the `mac` module instead.

brycx:
After discussing with a potential user, it's been decided that the current API, that offers keyed-mode for BLAKE2b in `hazardous`, will be kept as-is in terms of return-types.

PR #88, opened 2019-09-01
Change high-level auth to a wrapper around keyed Blake2b

vlmutolo:
Let me know if this is what you had in mind for a "wrapper" around keyed Blake2b. I think this is what we discussed, but obviously I'm happy to make changes.


vlmutolo:
This relates to #58. Instead of changing the return type, it was wrapped such that the output digest was replaced with a `Tag`.

src/auth.rs
```diff
@@ -61,30 +61,34 @@
...
+const BLAKE2B_TAG_SIZE: usize = 16;
 
 #[must_use]
-/// Authenticate a message using HMAC-SHA512.
+/// Authenticate a message using Blake2b in keyed mode.
 pub fn authenticate(secret_key: &SecretKey, data: &[u8]) -> Result<Tag, UnknownCryptoError> {
-	let mut state = hmac::init(&hmac::SecretKey::from_slice(
-		secret_key.unprotected_as_bytes(),
-	)?);
+	let blake2b_secret_key = blake2b::SecretKey::from_slice(secret_key.unprotected_as_bytes())?;
```
brycx: I think it would be worthwhile to enforce a 32-byte secret key minimum size. Could you please add a check to make sure the `secret_key` is greater-than or equal to 32 in length? If it is not, `UnknownCryptoError` needs to be returned. By enforcing 32-byte minimum, we also don't need to specify the lower length limitation of the `hazardous` secret key, which cannot be 0 or empty.

This restriction should be mentioned in the errors-section of the docs and have an accompanying unit-test.

src/auth.rs
```diff
@@ -45,8 +45,8 @@
 //! # Security:
```
brycx: The errors-section is missing a propagated error, because the `SecretKey` used in `hazardous` can max be 64 bytes for Blake2b. This limitation needs to be mentioned in the docs.

src/auth.rs
```diff
@@ -61,30 +61,34 @@
...
 
 #[must_use]
-/// Authenticate a message using HMAC-SHA512.
+/// Authenticate a message using Blake2b in keyed mode.
 pub fn authenticate(secret_key: &SecretKey, data: &[u8]) -> Result<Tag, UnknownCryptoError> {
-	let mut state = hmac::init(&hmac::SecretKey::from_slice(
-		secret_key.unprotected_as_bytes(),
-	)?);
+	let blake2b_secret_key = blake2b::SecretKey::from_slice(secret_key.unprotected_as_bytes())?;
+	let mut state = blake2b::init(Some(&blake2b_secret_key), BLAKE2B_TAG_SIZE)?;
```
brycx: Output size should be 32.

src/auth.rs
```diff
@@ -61,30 +61,34 @@
...
 #[must_use]
-/// Authenticate and verify a message using HMAC-SHA512.
+/// Authenticate and verify a message using Blake2b in keyed mode.
 pub fn authenticate_verify(
 	expected: &Tag,
 	secret_key: &SecretKey,
 	data: &[u8],
 ) -> Result<bool, UnknownCryptoError> {
-	let key = hmac::SecretKey::from_slice(secret_key.unprotected_as_bytes())?;
-	hmac::verify(expected, &key, data)
```
brycx: The same 32-byte minimum limitation on SecretKey should be checked here, as well as tested.

src/auth.rs
```diff
@@ -61,30 +61,34 @@
...
+	let key = blake2b::SecretKey::from_slice(secret_key.unprotected_as_bytes())?;
+	let expected_digest = blake2b::Digest::from_slice(expected.unprotected_as_bytes())?;
+	blake2b::verify(&expected_digest, &key, BLAKE2B_TAG_SIZE, data)
```
brycx: `BLAKE2B_TAG_SIZE` should be `32`.

src/hltypes.rs
```diff
@@ -72,6 +72,17 @@ construct_tag! {
 
 impl_from_trait!(PasswordHash, 128);
 
+construct_tag! {
+	/// A type to represent the `Tag` output by Blake2b hash in keyed mode.
+	///
+	/// # Errors:
+	/// An error will be returned if:
+	/// - `slice` is not 16 bytes.
+	(Blake2bTag, test_blake2b_tag, 16, 16)
```
brycx: 1) It should be named `Tag` since there should only be one tag used in high-level, that is in `auth`.
2) Please change the lower and upper bounds to be `32` and reflect the change in documentation, since we want to use `BLAKE2b-256`.

src/hltypes.rs
```diff
@@ -72,6 +72,17 @@ construct_tag! {
...
+}
+
+impl_from_trait!(Blake2bTag, 16);
```
brycx: Name `Tag` and `32`.

brycx:
For future reference: The reason that a 32-byte key and tag have been chosen is due to this being in line with what [libsodium recommends](https://download.libsodium.org/doc/hashing/generic_hashing) in its generic-hash interface, as well as [this answer](https://crypto.stackexchange.com/questions/57600/recommended-key-lengths-for-blake2b).

src/auth.rs
```diff
@@ -45,8 +45,8 @@
 //! # Security:
 //! - The secret key should always be generated using a CSPRNG.
 //!   [`SecretKey::default()`] can be used for
-//! this, it will generate a [`SecretKey`] of 32 bytes.
-//! - The recommended minimum length for a [`SecretKey`] is 32.
+//! this, it will generate a [`SecretKey`] of 16 bytes.
```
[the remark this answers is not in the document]
vlmutolo: Sorry, I thought the design had landed on 16-byte (128-bit). But 256-bit sounds good to me.

src/auth.rs
```diff
@@ -45,8 +45,8 @@
...
+//! - The recommended length for a [`SecretKey`] is 16 bytes.
```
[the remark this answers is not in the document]
vlmutolo: Ah, that makes sense. Gotcha.

src/auth.rs
vlmutolo: Oh, I thought that the internal call to `blake2b::SecretKey::from_slice` did the check automatically. But that doesn't do any checks. So we have to do one in this function. Does that sound right?

It also seems like we should return an `UnknownCryptoError` instead of panicking, right?

brycx: Exactly, but both in `authenticate()` and `authenticate_verify()`. 

Note: `blake2b::SecretKey::from_slice` does do a check, but checks that the slice is in range of `1..=64`.

src/auth.rs
```diff
@@ -87,9 +95,6 @@ pub fn authenticate_verify(
 	secret_key: &SecretKey,
 	data: &[u8],
 ) -> Result<bool, UnknownCryptoError> {
-	if secret_key.get_length() < BLAKE2B_MIN_KEY_SIZE {
-		return Err(UnknownCryptoError);
-	}
```
brycx: This length check should not have been removed. We want to check in both functions.

src/auth.rs
```diff
@@ -129,13 +134,11 @@ mod public {
...
 		#[test]
-		fn test_authenticate_verify_key_too_small() {
+		fn test_authenticate_key_too_small() {
 			let sec_key = SecretKey::generate(31).unwrap();
 			let msg = "what do ya want for nothing?".as_bytes().to_vec();
 
-			let mac_bob = authenticate(&sec_key, &msg).unwrap();
-
-			assert!(authenticate_verify(&mac_bob, &sec_key, &msg).is_err());
+			assert!(authenticate(&sec_key, &msg).is_err());
```
brycx: After the length check is added in `authenticate_verify()` we want to tests the corresponding error here too.

On top of that, we want to test that the length can actually be in range `32..=64` and that the hazardous `from_slice` error is correctly propagated. So we need to add tests for failure on `65` and pass on `64` as well. Ideally, you could tests this with a `quickcheck!` function. There are a lot of those used in the blake2b hazardous module.

src/auth.rs
```diff
@@ -42,6 +42,10 @@
 //! An error will be returned if:
 //! - The calculated [`Tag`] does not match the expected.
 //!
```
brycx: We should mention that an error will be returned if the secret key length is less than 32 or greater than 64.

src/auth.rs
brycx: It seems I was looking at outdated changes here, but the other comments seem valid.

brycx:
`libs.rs` L35-36 also needs to be updated. Currently it says the following:
```
//! ## Message authentication
//! [`orion::auth`] offers message authentication and verification using HMAC.
```

This should now state the use of BLAKE2b.

src/auth.rs
brycx: Fixed in https://github.com/brycx/orion/pull/88/commits/dbc722e105305e3d5669c0103d676e30d47e7dde

src/auth.rs
brycx: Fixed in https://github.com/brycx/orion/pull/88/commits/16d0e94a497661e16a74ce36d7edb48a358505df

Issue #90, opened 2019-09-03
Adding security-warnings to #[must_use]


brycx:
Functions that return a `Result` generate a warning when these are not used:
```
warning: unused `std::result::Result` that must be used
   --> src/high_level_api.rs:111:5
    |
111 |     orion::aead::seal(&aead_key, "Test".as_bytes());
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: #[warn(unused_must_use)] on by default
    = note: this `Result` may be an `Err` variant, which should be handled
```

This is the standard warning for a `Result`, but adding an explicit security warning to these functions may help users understand the importance of not ignoring a `Result`. Adding this would then produce two separate warnings by the compiler:

`#[must_use = "SECURITY WARNING: Ignoring a Result may have security-implications."]
`

```
warning: unused `std::result::Result` that must be used
   --> src/high_level_api.rs:111:5
    |
111 |     orion::aead::seal(&aead_key, "Test".as_bytes());
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: #[warn(unused_must_use)] on by default
    = note: this `Result` may be an `Err` variant, which should be handled

warning: unused return value of `orion::aead::seal` that must be used
   --> src/high_level_api.rs:111:5
    |
111 |     orion::aead::seal(&aead_key, "Test".as_bytes());
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: SECURITY WARNING: Ignoring a Result may have security-implications.
```

The question is whether this approach is worthwhile.

Edit: If this is decided against, then we should remove the useless `#[must_use]` on `Result` as mentioned in #89.

brycx:
I have gotten positive feedback on this and it will therefor be implemented. This is based on the same assumption as mentioned above: that users hopefully will be more cautious about ignoring such `Result`s which can have a security-impact. The error message that seemed most clear and concise was: 

`"SECURITY WARNING: Ignoring a Result can have real security implications."`

brycx:
@colelawrence expressed interest and is taking this on.

brycx:
Fixed in #95.

PR #95, opened 2019-09-06
Address #90: Add must_use message to pub Results

colelawrence:
+ Remove redundant must_use from internal fns

### DIFF
--- a/src/hazardous/hash/sha512.rs
+++ b/src/hazardous/hash/sha512.rs
@@ -389,7 +389,6 @@ impl Sha512 {
 	}
 }
 
-#[must_use = "SECURITY WARNING: Ignoring a Result can have real security implications."]
 /// Initialize a `Sha512` struct.
 pub fn init() -> Sha512 {
 	Sha512 {
--- a/src/hazardous/hash/sha512.rs
+++ b/src/hazardous/hash/sha512.rs
@@ -291,7 +291,7 @@ impl Sha512 {
 		}
 	}
 
-	#[must_use]
+	#[must_use = "SECURITY WARNING: Ignoring a Result can have real security implications."]
 	/// Update state with `data`. This can be called multiple times.
 	pub fn update(&mut self, data: &[u8]) -> Result<(), UnknownCryptoError> {
 		if self.is_finalized {
@@ -347,7 +347,7 @@ impl Sha512 {
 		Ok(())
 	}
 
-	#[must_use]
+	#[must_use = "SECURITY WARNING: Ignoring a Result can have real security implications."]
 	/// Return a SHA512 digest.
 	pub fn finalize(&mut self) -> Result<Digest, UnknownCryptoError> {
 		if self.is_finalized {
@@ -389,7 +389,7 @@ impl Sha512 {
 	}
 }
 
-#[must_use]
+#[must_use = "SECURITY WARNING: Ignoring a Result can have real security implications."]
 /// Initialize a `Sha512` struct.
 pub fn init() -> Sha512 {
 	Sha512 {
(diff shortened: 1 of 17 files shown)


src/hazardous/hash/sha512.rs
```diff
@@ -389,7 +389,7 @@ impl Sha512 {
 	}
 }
 
-#[must_use]
+#[must_use = "SECURITY WARNING: Ignoring a Result can have real security implications."]
 /// Initialize a `Sha512` struct.
 pub fn init() -> Sha512 {
```
brycx: This function does not return a `Result`, so the use of `#[must_use]` should be removed.

src/hazardous/mac/hmac.rs
```diff
@@ -94,7 +94,7 @@ construct_tag! {
 
 impl_from_trait!(Tag, SHA512_OUTSIZE);
 
-#[must_use]
+#[must_use = "SECURITY WARNING: Ignoring a Result can have real security implications."]
 #[derive(Clone)]
```
brycx: This state should have its `#[must_use]` removed I believe.

src/hazardous/mac/poly1305.rs
```diff
@@ -399,7 +398,7 @@ impl Poly1305 {
 	}
 }
 
-#[must_use]
+#[must_use = "SECURITY WARNING: Ignoring a Result can have real security implications."]
 /// Initialize a `Poly1305` struct with a given one-time key.
 pub fn init(one_time_key: &OneTimeKey) -> Poly1305 {
```
brycx: This function does not return a `Result`, so the use of `#[must_use]` should be removed.

src/hazardous/stream/chacha20.rs
```diff
@@ -307,7 +306,6 @@ impl InternalState {
 		})
 	}
```
brycx: Both `encrypt()`, `decrypt()` and `keystream_block()` are missing the new `#[must_use]` warnings. Whereas `hchacha20()` and `encrypt_in_place` should have theirs removed.

brycx:
Closes #90.

brycx:
@colelawrence Do you know if it's possible to specify the warning message as a `&'static str` and have it saved in the `errors` module? If it is, I think that would be a better approach than having to edit every single one, in case we want to change it later.

colelawrence:
> @colelawrence Do you know if it's possible to specify the warning message as a `&'static str` and have it saved in the `errors` module? If it is, I think that would be a better approach than having to edit every single one, in case we want to change it later.

I'll look into that and implement your suggestions

src/hazardous/stream/chacha20.rs
colelawrence: For this one, should we remove must_use because the docs are hidden?
```rs
#[must_use]
#[doc(hidden)]
/// HChaCha20 as specified in the [draft-RFC](https://github.com/bikeshedders/xchacha-rfc/blob/master).
pub fn hchacha20(
	secret_key: &SecretKey,
	nonce: &[u8],
) -> Result<[u8; HCHACHA_OUTSIZE], UnknownCryptoError> {
```

brycx: Yes, precisely.

colelawrence:
I haven't been running `cargo fmt`. I didn't know it was available.

I ran cargo fmt and [this happened](https://github.com/brycx/orion/pull/95/commits/6fe27727861ae33ba8d8a78bb353f833aa07d5df) is that normal? Sorry, if there's something silly I'm missing...

brycx:
> I ran cargo fmt and this happened is that normal? Sorry, if there's something silly I'm missing...

I think it's some mis-formatting in rustfmt. But since we check that in CI, this is how it's "supposed" to look.

colelawrence:
Iasked about the static string idea, and I heard that we won't be able to use a static str or const within the attribute macro syntax due to different compilation steps.

However, it would be possible to create a `#[orion_must_use]` proc-macro for generating the same `#[must_use=message]`

I would guess it could be relatively simple, but I haven't looked into it beyond that.

brycx:
> I asked about the static string idea, and I heard that we won't be able to use a static str or const within the attribute macro syntax due to different compilation steps.

Thanks for looking into this. I think we can just keep this as-is for now. There aren't that many messages after all and I feel like using proc-macros would add unneeded complexity to something that is fairly simple.

Issue #94, opened 2019-09-06
Stream encryption API


snsmac:
Something that i would like to see in this library is an API for encrypting a stream of messages.
This is for example useful if a file is too large to be encrypted at once.

It is impotent for such an API that messages can not be modified , dropped or reordered. This means the complete stream needs to be authenticated. 
Its also a requirement, that the same message, if encrypted repeatedly, produces different ciphertexts.

Libsodium provides such a functionality based on XChaCha20Poly1305. 
https://download.libsodium.org/doc/secret-key_cryptography/secretstream

This functionality can also easily implemented for orion. (I already have a working, hazardous implementation for it )

The main problem that i see is that this algorithm isn't standardized. 

If you want this functionality  i can submit a pull request for this

brycx:
> Something that i would like to see in this library is an API for encrypting a stream of messages.
This is for example useful if a file is too large to be encrypted at once.

This is a great idea! I'm curious whether you plan on using such a feature anywhere specific?

> The main problem that i see is that this algorithm isn't standardized.

XChaCha20Poly1305 is in the process of being standardised and is used quite a few other places as well. I see no issue with basing a streaming encryption API upon this. However, we could just as well add support for streaming ChaCha20Poly1305 encryption, should the non-standardisation cause concern to anyone.

The libsodium secretstream is a good baseline for implementing it in orion.

> If you want this functionality i can submit a pull request for this

Yes, this definitely seems worthwhile to provide. I have looked into it before, but not thought about specific requirements for such an implementation, which means we might need to re-iterate over its design a few times. Since you mentioned you already have a working implementation, you could open it as WIP while we work through it?

PR #99, opened 2019-09-09
XChaCha20Poly1305 streaming encryption

snsmac:
This is the "hazardous" part of the secreat stream API proposed in #94.

Things that are still missing are:

- Documentation
- More tests, especially with wrong inputs
- An user friendly, "safe" API


brycx:
Thank you for putting this together @snsmac! I don't have the time atm, but I will dive into the code more deeply in the coming days.

A preliminary observation is that you're pulling in another dependency `bitflags`. It's always preferable to avoid this, do you think this could be implemented without that extra dependency?

src/hazardous/secret_stream/xchacha20poly1305.rs
```diff
@@ -0,0 +1,455 @@
...
+		}
+
+		chacha20_enc_in_place(&self.get_key(), &self.get_nonce(), 0, &mut block)?;
+		let mut poly = init(&OneTimeKey::from_slice(&block[..POLY1305_KEYSIZE])?);
+		use zeroize::Zeroize;
+		block.zeroize();
+		if adlen > 0 {
+			poly.update(ad.unwrap())?;
+		}
+		poly.update(&pad[..((16usize.overflowing_sub(adlen).0) & 15)])?;
```
brycx: There is already a padding function available in `chacha20poly1305`. If we make that public (`pub(crate)`), then we could use it like the other AEAD implementations do.

src/hazardous/secret_stream/xchacha20poly1305.rs
```diff
@@ -0,0 +1,455 @@
...
+		)?;
+		poly.update(&dst_out[cipherpos..(cipherpos + msglen)])?;
+		poly.update(
+			&pad[..((16usize
+				.overflowing_sub(BLOCKSIZE)
+				.0
+				.overflowing_add(msglen)
+				.0) & 15)],
+		)?;
+		store_u64_into_le(&[adlen as u64], &mut slen);
```
brycx: Instead of using `store_u64_into_le`(requires a few more allocations IIRC) there is `slen.copy_from_slice((adlen as u64).to_le_bytes())`.

src/hazardous/secret_stream/xchacha20poly1305.rs
```diff
@@ -0,0 +1,455 @@
...
+		let adlen = match ad {
+			Some(v) => v.len(),
+			None => 0,
+		};
+		let msgpos = core::mem::size_of::<Tag>();
+		let macpos = msgpos + mlen;
+		if plaintext_out.len() < mlen {
+			return Err(UnknownCryptoError);
+		}
+		chacha20_enc_in_place(&self.get_key(), &self.get_nonce(), 0, &mut block)?;
```
brycx: We can make `poly1305_key_gen` `pub(crate)` in `chacha20poly1305`, so that we avoid re-defining the key-generation multiple times.

src/hazardous/secret_stream/xchacha20poly1305.rs
```diff
@@ -0,0 +1,455 @@
...
+	/// TODO: Doc
+	/// Decryptes a message
+	pub fn pull(
+		&mut self,
+		cipher: &[u8],
+		ad: Option<&[u8]>,
+		plaintext_out: &mut [u8],
+		tag_out: &mut Tag,
+	) -> Result<(), UnknownCryptoError> {
+		if cipher.len() < SECRETSTREAM_XCHACHA20POLY1305_ABYTES {
```
brycx: If we check for `cipher.len() <= SECRETSTREAM_XCHACHA20POLY1305_ABYTES` instead, we also avoid processing empty ciphertexts.

src/hazardous/secret_stream/xchacha20poly1305.rs
```diff
@@ -0,0 +1,455 @@
...
+		poly.update(&dst_out[cipherpos..(cipherpos + msglen)])?;
+		poly.update(
+			&pad[..((16usize
+				.overflowing_sub(BLOCKSIZE)
+				.0
+				.overflowing_add(msglen)
+				.0) & 15)],
+		)?;
+		store_u64_into_le(&[adlen as u64], &mut slen);
+		poly.update(&slen)?;
```
brycx: We can split up the authentication into a separate function, like it has been done in `chacha20poly1305::process_authentication`, to avoid code-duplication and make it easier to review.

src/hazardous/stream/chacha20.rs
```diff
@@ -401,7 +401,7 @@ impl Serialize {
 }
 
 /// In-place IETF ChaCha20 encryption as specified in the [RFC 8439](https://tools.ietf.org/html/rfc8439).
-fn encrypt_in_place(
+pub fn encrypt_in_place(
```
brycx: This should be `pub(crate)`.

src/typedefs.rs
```diff
@@ -23,6 +23,10 @@
 ///
 /// Trait implementation macros
 
+//TODO: Should the tag of a mac implement zeroize?
+// automatic zeroize on drop?
+// Should they implement Copy, Clone?
+
```
brycx: The tag isn't a secret like encryption keys are, so AFAICT there's no need to zeroize on drop. But `Copy`/`Clone` may be useful.

src/hazardous/secret_stream/xchacha20poly1305.rs
```diff
@@ -0,0 +1,455 @@
...
+//TODO assert that counterbytes + inoncebytes = IETF_CHACHA_NONCESIZE
+
+fn xor_buf8(out: &mut [u8], input: &[u8]) {
+	for i in 0..8 {
+		out[i] ^= input[i];
+	}
+}
+
+fn increment4(buf: &mut [u8]) {
+	let mut c = 1u16;
```
brycx: We can add a pre-condition here that buf must have a lest 4 elements. I think `increment_nonce` would be a more clear name to the function.

src/hazardous/secret_stream/xchacha20poly1305.rs
```diff
@@ -0,0 +1,455 @@
...
+		let mut slen = [0u8; 8];
+		let pad = [0u8; 16];
+		let adlen = match ad {
+			Some(v) => v.len(),
+			None => 0,
+		};
+		let msglen = plaintext.len();
+		let cipherpos = core::mem::size_of::<Tag>();
+		let macpos = cipherpos + msglen;
+		if dst_out.len() < SECRETSTREAM_XCHACHA20POLY1305_ABYTES + msglen {
```
brycx: We should make this length check before we allocate the buffers.

src/hazardous/secret_stream/xchacha20poly1305.rs
```diff
@@ -0,0 +1,455 @@
...
+				.0
+				.overflowing_add(msglen)
+				.0) & 15)],
+		)?;
+		store_u64_into_le(&[adlen as u64], &mut slen);
+		poly.update(&slen)?;
+		store_u64_into_le(&[BLOCKSIZE as u64 + msglen as u64], &mut slen);
+		poly.update(&slen)?;
+		let mac = poly.finalize()?;
+		dst_out[macpos..(macpos + mac.get_length())].copy_from_slice(mac.unprotected_as_bytes());
```
brycx: Would be nice to have a `debug_assert` to check that `dst_out` actually has `macpos + mac.get_length()` elements.

src/hazardous/secret_stream/xchacha20poly1305.rs
```diff
@@ -0,0 +1,455 @@
...
+#[cfg(test)]
+mod private {
+	extern crate hex;
+	use self::hex::decode;
+	use super::*;
+	use hex::encode;
+
+	//All test values generated with libsodium https://github.com/jedisct1/libsodium version 1.0.18
+
+	#[test]
```
brycx: This is great! It would be awesome if you could include the script/program that was used to create the test-vectors in `/tests/test_generation`.

src/hazardous/secret_stream/xchacha20poly1305.rs
```diff
@@ -0,0 +1,455 @@
...
+use crate::util::secure_cmp;
+use bitflags::bitflags;
+
+//TODO Add more tests
+
+//TODO: Im not sure if this is the right place for the bit flags and if they are needed at all.
+// It would also be possible to only define the REKEY value and pass the tag around as u8 and define
+// the user friendly bitflags/enum in the safe API. Im not sure how to work around using the
+// bitflags create. rust enums are rather limited in functionality as flags and im not a fan of
+// passing around raw u8 bitflags.
```
brycx: I think this is the right place to put it. In regards to using the bitflags crate, I think we can keep it for now. What is really nice is that it has no `unsafe` usage and is used in a number of other crates as well.

src/hazardous/aead/chacha20poly1305.rs
```diff
@@ -192,15 +196,39 @@ pub fn seal(
...
+		optional_ad,
+		&crypt_out,
+		plaintext.len(),
+	)?;
+	mac_out.copy_from_slice(poly1305_state.finalize()?.unprotected_as_bytes());
+
+	Ok(())
+}
+
+#[must_use]
```
brycx: NIT: We need the new warning message here as well.

src/hazardous/secret_stream/xchacha20poly1305.rs
```diff
@@ -0,0 +1,455 @@
...
+	/// prepare a new push or pull operation
+	pub fn init(&mut self, key: chacha20Key, nonce: Nonce) {
+		*self = SecretStreamXChaCha20Poly1305::new(key, nonce);
+	}
+
+	#[must_use = "SECURITY WARNING: Ignoring a Result can have real security implications."]
+	/// derives a new secret key used for encryption/decryption
+	pub fn rekey(&mut self) -> Result<(), UnknownCryptoError> {
+		let mut new_key_and_inonce =
+			[0u8; CHACHA_KEYSIZE + SECRETSTREAM_XCHACHA20POLY1305_INONCEBYTES]; //TODO
```
brycx: Is there anything missing, since there is a `TODO`?

src/hazardous/secret_stream/xchacha20poly1305.rs
```diff
@@ -0,0 +1,455 @@
...
+	/// TODO: Doc
+	/// //Ecryptes a message
+	pub fn push(
+		&mut self,
+		plaintext: &[u8],
+		ad: Option<&[u8]>,
+		dst_out: &mut [u8],
+		tag: Tag,
+	) -> Result<(), UnknownCryptoError> {
+		const BLOCKSIZE: usize = 64;
```
brycx: This blocksize should be defined in the module-root, since both pull and push use it.

src/hazardous/aead/chacha20poly1305.rs
```diff
@@ -168,14 +168,18 @@ fn process_authentication(
 
 #[must_use = "SECURITY WARNING: Ignoring a Result can have real security implications."]
 /// AEAD ChaCha20Poly1305 encryption and authentication as specified in the [RFC 8439](https://tools.ietf.org/html/rfc8439).
-pub fn seal(
+pub fn seal_split(
```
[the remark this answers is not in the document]
snsmac: I played around with a different implementation before and forgot to remove this change. I reverted it now

src/hazardous/secret_stream/xchacha20poly1305.rs
```diff
@@ -0,0 +1,455 @@
...
+	/// An error will be returned if:
+	/// - `slice` is not 24 bytes.
+	(Nonce, test_nonce, SECRETSTREAM_XCHACHA20POLY1305_HEADERBYTES, SECRETSTREAM_XCHACHA20POLY1305_HEADERBYTES)
+}
+
+impl_from_trait!(Nonce, SECRETSTREAM_XCHACHA20POLY1305_HEADERBYTES);
+
+/// Secret Stream State
+pub struct SecretStreamXChaCha20Poly1305 {
+	key: [u8; CHACHA_KEYSIZE],
```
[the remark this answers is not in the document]
snsmac: Not sure whats the best approach here. It is possible to reuse the chacha20 SecretKey here, but then i would need to call `unprotected_as_bytes()` whenever it is used

src/hazardous/secret_stream/xchacha20poly1305.rs
```diff
@@ -0,0 +1,454 @@
...
+/// Size of additional data appended to each message
+pub const SECRETSTREAM_XCHACHA20POLY1305_ABYTES: usize =
+	POLY1305_OUTSIZE + core::mem::size_of::<Tag>();
+//TODO assert that counterbytes + inoncebytes = IETF_CHACHA_NONCESIZE
+
+fn xor_buf8(out: &mut [u8], input: &[u8]) {
+	debug_assert_eq!(out.len(), 8);
+	debug_assert_eq!(input.len(), 8);
+	for i in out.iter_mut().zip(input.iter()) {
+		*i.0 ^= *i.1;
```
brycx: NIT: Instead of combining these two I think the loop could be made clearer by splitting them up:
```
for (o_elem, i_elem) in out.iter_mut().zip(input.iter()) {
		*o_elem ^= i_elem;
```

src/hazardous/secret_stream/xchacha20poly1305.rs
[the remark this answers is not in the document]
brycx: It seems most calls to the key are in the form of `get_key()`, which converts the array to a `SecretKey`. If the only concern is that it adds the additional function call in some places, seems to be mostly the re-keying, then I think we should definitely move towards the struct owning a `SecretKey`.

src/hazardous/secret_stream/xchacha20poly1305.rs
brycx: Do you happen to have a use-case that could help me understand when you'd want to send only tags that carry information and when encrypting streams? That is, scenarios where any such information can not be attached to a message?

src/hazardous/secret_stream/xchacha20poly1305.rs
brycx: It's now available on master.

src/hazardous/secret_stream/xchacha20poly1305.rs
brycx: This is now available on master.

snsmac:
This changes should resolve most of the raised issues (tests and documentation are still missing).
There are still a few things that I'm unsure about:

-  Should additional authenticated data passed as slice or as an Option ?
- I still zero the mac on a failed comparion. (https://github.com/snsmac/orion/blob/e613461e2aca63f8fcb7b1b894dd269ff9566c7e/src/hazardous/secret_stream/xchacha20poly1305.rs#L230) . I do not think this is necessary, but for some reason the libsodium implementation does it too.
- I do not understand your comment regarding the tag_out parameter (https://github.com/snsmac/orion/blob/e613461e2aca63f8fcb7b1b894dd269ff9566c7e/src/hazardous/secret_stream/xchacha20poly1305.rs#L207) This value is needed by the decrypting party to know if it can expect more messages

src/hazardous/secret_stream/xchacha20poly1305.rs
snsmac: I think it should be generally possible to attach the tag to an message.

The question if empty ciphertexts should be allowed is more a generally design decision for this library. The rejection of empty texts can catch some user errors, but personally i would do this in the safe API and allow them in the hazardous API. This allows users that for some reasons want to do some operation on an empty text to do so.   Allowing empty texts should generally have no impact on memory or cryptographic safety.

brycx:
> Should additional authenticated data passed as slice or as an Option ?

Let's pass it as an `Option<&[u8]>` to keep it consistent with the other interfaces.

> I do not think this is necessary, but for some reason the libsodium implementation does it too.

I don't think it's necessary either, but I suspect that it's just the default way to free/clean variables in libsodium. We should be able to remove it for the time being.

> I do not understand your comment regarding the tag_out parameter 

Sorry, let me try to clarify. The tag is encrypted and included in the message AFAIU. Because the tag is included in the encrypted message, I don't see why we need to store it additionally in an out parameter. In case we don't actually have to, then we should just call it `tag` instead of `tag_out`.

I also think it would make sense to add a `finalized` field to the stream encryption state, so it checks at every operation whether or not a `Final` message has already been encrypted. If it has, it should return an error and force the user to reset the stream.

src/hazardous/secret_stream/xchacha20poly1305.rs
```diff
@@ -0,0 +1,428 @@
...
+	/// TODO: Doc
+	/// Decryptes a message
+	pub fn decrypt_message(
+		&mut self,
+		cipher: &[u8],
+		ad: &[u8],
+		plaintext_out: &mut [u8],
+		tag_out: &mut Tag,
+	) -> Result<(), UnknownCryptoError> {
+		let msglen = cipher.len() - SECRETSTREAM_XCHACHA20POLY1305_ABYTES;
```
brycx: The length check must come before this subtraction, otherwise it will silently overflow in release builds.

snsmac:
> Sorry, let me try to clarify. The tag is encrypted and included in the message AFAIU. Because the tag is included in the encrypted message, I don't see why we need to store it additionally in an out parameter. In case we don't actually have to, then we should just call it `tag` instead of `tag_out`.

Sorry, I still don't really understand you.
Maybe the following pseudocode for decrpyting a stream helps to understand why there is the tag_out field.
```
while(tag != Tag::Finish) 
{
   // get ciphertext
   decrypt_message(&ciphertext, &mut tag, &mut plaintext);
  //do something with plaintext
}
```

So the tag is needed so the decrypting code knows when the stream is finished.

I will look at the proposed changes and tests later, but this week i have no time for this

brycx:
> Maybe the following pseudocode for decrpyting a stream helps to understand why there is the tag_out field.

Yes, this makes it clearer to me. I think I was the one to misunderstand something, thanks for clarifying. 

So the tag parameter needs to be returned, though I still think we can avoid it to be a mutable parameter in the decrypting function call. I want to avoid these out parameters when we can, since they impact usability. Currently a new tag is already allocated in the decryption function, so we could just return that instead.

I've also started looking into tests to see if I can design some cases for integration in the new test framework (#96). I will also begin sketching some fuzzing targets. The most important thing though are the test vectors generated with libsodium. If you'd like to work on either one of those, let me know.

brycx:
I've set up a basic fuzzing target for encryption/decryption with `Tag::MESSAGE` and `Tag::FINISH`: https://github.com/brycx/orion-fuzz/tree/secret-stream

So far I've only been running it for 30 minutes, which hasn't uncovered any issues yet.

It seems setting up comparison with sodiumoxide is not possible, because its API does not allow to set or get the nonce used.

src/hazardous/aead/chacha20poly1305.rs
```diff
@@ -686,7 +686,7 @@ mod private {
 				// The usize that padding() returns should always
 				// be what remains to make input a multiple of 16 in length.
 				fn prop_padding_result(input: Vec<u8>) -> bool {
```
brycx: NIT: Now that padding takes `usize` we can have quickcheck generate those instead of `input: Vec<u8>`: `input: usize`.

src/hazardous/secret_stream/xchacha20poly1305.rs
```diff
@@ -0,0 +1,888 @@
...
+		let msgpos = core::mem::size_of::<Tag>();
+		let macpos = msgpos + msglen;
+
+		block[0] = ciphertext[0];
+		chacha20_enc_in_place(&self.key, &self.get_nonce(), 1, &mut block)?;
+		let tag = Tag::from_bits(block[0]).ok_or(UnknownCryptoError)?;
+		block[0] = ciphertext[0];
+		let mac = self
+			.generate_auth_tag(ciphertext, ad, msglen, &block, adlen, msgpos)
+			.unwrap();
```
brycx: Please propagate with `?`.

src/hazardous/secret_stream/xchacha20poly1305.rs
```diff
@@ -0,0 +1,888 @@
...
+			&self.key,
+			&self.get_nonce(),
+			2,
+			plaintext,
+			&mut dst_out[cipherpos..],
+		)?;
+
+		let mac = self
+			.generate_auth_tag(dst_out, ad, msglen, &block, adlen, cipherpos)
+			.unwrap();
```
brycx: Please propagate with `?`.

src/hazardous/secret_stream/xchacha20poly1305.rs
brycx: I would like to continue rejecting empty input for now at least. So we have to skip the plaintext encryption call if that is empty, but still authenticate the block that is then put in `dst_out`. Does that make any sense?

src/typedefs.rs
brycx: Fixed in https://github.com/brycx/orion/pull/99/commits/c962e9b5124307bf108a4dfb3b6a858128ca24f3

src/hazardous/secret_stream/xchacha20poly1305.rs
brycx: Fixed in https://github.com/brycx/orion/pull/99/commits/c962e9b5124307bf108a4dfb3b6a858128ca24f3

src/hazardous/secret_stream/xchacha20poly1305.rs
```diff
@@ -0,0 +1,888 @@
...
+		};
+		let adlen = ad.len();
+		let cipherpos = core::mem::size_of::<Tag>();
+		let macpos = cipherpos + msglen;
+
+		block[0] = tag.bits();
+		chacha20_enc_in_place(&self.key, &self.get_nonce(), 1, &mut block)?;
+		dst_out[0] = block[0];
+
+		chacha20_enc(
```
brycx: We have to skip on empty `plaintext` here.

src/hazardous/secret_stream/xchacha20poly1305.rs
```diff
@@ -0,0 +1,888 @@
...
+		chacha20_enc_in_place(&self.key, &self.get_nonce(), 1, &mut block)?;
+		let tag = Tag::from_bits(block[0]).ok_or(UnknownCryptoError)?;
+		block[0] = ciphertext[0];
+		let mac = self
+			.generate_auth_tag(ciphertext, ad, msglen, &block, adlen, msgpos)
+			.unwrap();
+		if !(mac == &ciphertext[macpos..macpos + mac.get_length()]) {
+			return Err(UnknownCryptoError);
+		}
+		chacha20_enc(
```
brycx: We have to skip on empty `ciphertext` here.

brycx:
There was a problem with caching in Travis CI and the crate used to run code coverage, which caused the builds to fail. If you rebase to master this should be fixed now.

brycx:
Proposal for high-level interface: #103.

snsmac:
The CI test with sanitizers still times out

brycx:
Clearing the cache for the build seemed to do the trick, should be fixed now.

brycx:
I pushed some NITs that I didn't feel you should waste your time on. The only "big" change was to change `encrypt_message` and `decrypt_message` to `seal_chunk` and `open_chunk` for consistency.

src/hazardous/aead/xchacha20poly1305_stream.rs
```diff
@@ -328,20 +328,23 @@ impl SecretStreamXChaCha20Poly1305 {
...
 		poly.update(&pad[..padding(CHACHA_BLOCKSIZE.wrapping_sub(msglen))])?;
-		slen.copy_from_slice(&(adlen as u64).to_le_bytes());
-		poly.update(&slen)?;
-		slen.copy_from_slice(&(CHACHA_BLOCKSIZE as u64 + msglen as u64).to_le_bytes());
-		poly.update(&slen)?;
+		pad[..8].copy_from_slice(&(adlen as u64).to_le_bytes());
+		// TODO: How likely is it for users to trigger panic on this .unwrap()?
+		pad[8..16]
+			.copy_from_slice(&(CHACHA_BLOCKSIZE.checked_add(msglen).unwrap() as u64).to_le_bytes());
+		poly.update(&pad)?;
```
snsmac: The only way to trigger an overflow here is if usize is smaller then 64 bits and if msglen + 64 does not fit into usize.
This seems to be rather unlikely, but it can happen. With casting to u64 before the addition this code is safe. The performance impact of this cast should be minimal.

src/hazardous/aead/xchacha20poly1305_stream.rs
```diff
@@ -233,6 +234,14 @@ impl SecretStreamXChaCha20Poly1305 {
 		dst_out: &mut [u8],
 		tag: Tag,
 	) -> Result<(), UnknownCryptoError> {
+		// The largest addition done on length
+		// of plaintext. Other non-checked additions
+		// should not be able to overflow if this
+		// does not panic.
+		assert!((CHACHA_BLOCKSIZE as u64)
+			.checked_add(plaintext.len() as u64)
+			.is_some());
```
snsmac: I don't think this asserts are even necessary.
If i understand it correctly chach20 panics if it encrypts more then 2^32-1 * 64 bytes.  This is less then the message length where a overflow in the addition (if done on u64) begins to happen.

brycx: You're right, this is not the right approach. This makes no sense in `seal_chunk` because the addition we're checking for comes after the call to ChaCha20. However, the preceding additions, on 64-bit targets, would still be able to overflow it seems. Specifically L246 and L256. 

What if we instead in `seal_chunk` checked for `assert!(SECRETSTREAM_XCHACHA20POLY1305_ABYTES.checked_add(msglen).is_some())`, because this is the most added to `msglen` without casting to `u64`. If I'm not mistaken, this way we can be sure that the remaining additions done in `seal_chunk` and subsequently in `generate_auth_tag` will not overflow, both on 32 and 64-bit targets. What do you think?

src/hazardous/aead/xchacha20poly1305_stream.rs
```diff
@@ -0,0 +1,1361 @@
...
+		debug_assert!(dst_out.len() >= macpos + mac.get_length());
+		dst_out[macpos..(macpos + POLY1305_OUTSIZE)].copy_from_slice(mac.unprotected_as_bytes());
+
+		xor_slices_8(
+			self.inonce.as_mut(),
+			&dst_out[macpos..macpos + SECRETSTREAM_XCHACHA20POLY1305_INONCEBYTES],
+		);
+		self.counter = self.counter.wrapping_add(1);
+
+		if bool::from(tag.bits().ct_eq(&Tag::REKEY.bits()) | self.counter.ct_eq(&0u32)) {
```
brycx: This check will never trigger a rekey on `Tag::FINISH` due to comparison on its bits instead of using bitwise-AND. The new tests I've added should also detect this.

This has to be changed to something in the lines of:
```
bool::from(!(tag.bits() & Tag::REKEY.bits()).ct_eq(&0u8) | self.counter.ct_eq(&0u32))
```

Similarly in `open_chunk`.

src/hazardous/aead/xchacha20poly1305_stream.rs
snsmac: Do you prefer asserts or checked_add + unwrap for catching overflows ?
checked_add seems to generate a bit better assembly https://gcc.godbolt.org/z/Nec_kf

brycx: I like this, it's cleaner than what my approach would've ended up with. Feel free to add these changes.

brycx:
I've removed the WIP label as the PR is pretty much ready to be merged. I've looked through it again and haven't found anything functional that needed change. The fuzzing targets are also at a pretty stable point. I'll review this PR one last time within the next few days, and then we'll merge it. 

After that we can implement the high-level API in #103. Let me know if that's something you want to do @snsmac.

snsmac:
Before merging this i would prefer it to squash the commits together into one to have a cleaner commit history.
Implementing the high level API is something i can do

brycx:
I ended up adding more than I anticipated. The most notable was:

- `bitflags` crate has been replaced with a custom enum. `bitflags` provided a lot of unneeded functionality for the tag, which felt bloated for our use-case. This way we also avoid an extra dependency.
- `static_assertions` has been completely removed. This is mainly because the assertions were done on primitive types that have [stable size](https://doc.rust-lang.org/nightly/core/mem/fn.size_of.html) across different platforms, and can therefor be avoided. I said we should keep `static_assertions` because we could also use it to check trait implementations for newtypes. This is still true however, after looking into it further, this can be achieved rather easily without the extra dependency.
- Some general renaming to avoid long names.
- Added one test to test valid decryptable but incorrect stream tags. The one we already had would result in a invalidly decrypted tag.

If you agree with these changes, I can go ahead and merge the PR. Also, if you feel like it, take a look at the fuzzing target (https://github.com/brycx/orion-fuzz/pull/2/files#diff-f696f5056dafd1dcf872b75919e0bc0e). Definitely wouldn't hurt if someone besides me took a look at it.

src/hazardous/aead/streaming.rs
```diff
@@ -0,0 +1,1390 @@
...
+	fn eq(&self, other: &StreamTag) -> bool {
+		(self.as_byte().ct_eq(&other.as_byte())).into()
+	}
+}
+
+/// The size of the internal counter.
+const COUNTERBYTES: usize = 4;
+/// The size of the internal nonce.
+const INONCEBYTES: usize = 8;
+/// The size of a Stream.
```
snsmac: The comment here is wrong, its the size of a Tag not Stream

snsmac:
Generally it looks good to me, the fuzzing test also seems to be useful. 
Personally i prefer longer, descriptive names, but i can understand the preference for short names.
Do you want to squash the (58) commits together before merging into master ?

src/hazardous/aead/streaming.rs
brycx: Good catch.

brycx:
> Do you want to squash the (58) commits together before merging into master ?

Sorry, yeah. All commits are usually squashed from a PR here on GitHub when merged to master.

PR #108, opened 2019-10-30
Add high-level secret-stream API

### LINKED ISSUE #94: Stream encryption API

### LINKED ISSUE #103: Proposal for high-level "secret-stream" API
This is a proposal for the high-level interface, providing stream-based encryption based on the hazardous implementation in #99. I find the interface provided by [sodiumoxide](https://github.com/sodiumoxide/sodiumoxide) quite nice and my proposal borrows a fair amount from there for that reason.

The streaming encryption will be a part of the `orion::aead` module, defining two types. One for handling stream encryption (`StreamSealer`) and one to handle stream decryption (`StreamOpener`). Nonce generation will be done automatically. Providing separate types for encryption/decryption is to make misuse harder and is also a concern @snsmac has brought up. Here is a rough implementation of what this interface may look like:

```
pub struct StreamSealer {
	internal_sealer: SecretStreamXChaCha20Poly1305,
	nonce: Nonce
}

impl StreamSealer {

	pub fn get_nonce(&self) -> Nonce {
		self.nonce
	}

	pub fn new(secret_key: &SecretKey) -> Result<Self, UnknownCryptoError> {
		let nonce = Nonce::generate();
		let sk = &chacha20::SecretKey::from_slice(secret_key.unprotected_as_bytes())?;
		
		Ok(Self {
			internal_sealer: SecretStreamXChaCha20Poly1305::new(&sk, &nonce),
			nonce
		})
	}

	pub fn seal_chunk(&mut self, data: &[u8], ad: Option<&[u8]>, tag: ChunkTag) -> Result<Vec<u8>, UnknownCryptoError> {
		let sealed_chunk_len = data.len().checked_add(SECRETSTREAM_XCHACHA20POLY1305_ABYTES);
		if sealed_chunk_len.is_none() {
			return Err(UnknownCryptoError);
		}
		
		let mut sealed_chunk = vec![0u8; sealed_chunk_len.unwrap()];
		self.internal_sealer.encrypt_message(data, ad, &mut sealed_chunk, tag)?;

		Ok(sealed_chunk)	
	}
}

pub struct StreamOpener {
	internal_sealer: SecretStreamXChaCha20Poly1305,
}

impl StreamOpener {

	pub fn new(secret_key: &SecretKey, nonce: &Nonce) -> Result<Self, UnknownCryptoError> {
		let sk = &chacha20::SecretKey::from_slice(secret_key.unprotected_as_bytes())?;
		
		Ok(Self {
			internal_sealer: SecretStreamXChaCha20Poly1305::new(&sk, &nonce),
		})
	}


	pub fn open_chunk(&mut self, data: &[u8], ad: Option<&[u8]>) -> Result<(Vec<u8>, ChunkTag), UnknownCryptoError> {
		if data.len() < SECRETSTREAM_XCHACHA20POLY1305_ABYTES {
			return Err(UnknownCryptoError);
		}
		
		let mut opened_chunk = vec![0u8; data.len() - SECRETSTREAM_XCHACHA20POLY1305_ABYTES];
		let tag = self.internal_sealer.decrypt_message(data, ad, &mut opened_chunk)?;

		Ok((opened_chunk, tag))
	}
}
```

I specifically try to move away from calling the encryption/decryption operations `push` or `pull`, as I find these to be bad names as they don't clearly communicate the purpose or functionality behind IMO. Including "seal" and "open" was to be consistent with naming of AEAD-related functions throughout the rest of orion.

Once #88 is finalised, there will be a high-level type called `Tag` as well as all other `Tag`s used in hazardous. That is the reasoning behind importing the stream `Tag` as `ChunkTag`, though I don't know if there's something better to call it.

Another remaining question is whether the nonce used should be called a `Header` as libsodium does it, or if we should stick with `Nonce`. I personally find `Nonce` to be a more fitting name, because I find it to convey it's use more clearly. 

Any feedback/input is greatly appreciated!

snsmac:
Closes #103 and closes #94 

There is one link in the documentation that i can not get to work, i have marked it with a TODO.

I'm a bit unsure if it is better to have a get_nonce method or if it is better to return Self and the Nonce in the `new` method.

Tests a rather minimal for now, but i think this should be fine since its only a thin wrapper around the hazardous API

Is it useful to expose the option to authenticate additional data? The "normal" AEAD seal function does not do it

### DIFF
--- a/src/aead.rs
+++ b/src/aead.rs
@@ -173,6 +173,7 @@ pub fn open(
 ///
 /// # Errors:
 /// An error will be returned if:
+/// - `secret_key` is not 32 bytes.
 /// - The length of `ciphertext` is not greater than `16`.
 /// - The received mac does not match the calculated mac when decrypting. This can indicate
 ///   a dropped or reordered message within the stream.
@@ -185,6 +186,8 @@ pub fn open(
 /// - Failure to generate random bytes securely.
 ///
 /// # Security:
+/// - It is critical for security that a given nonce is not re-used with a given
+///   key.
 /// - To securely generate a strong key, use [`SecretKey::generate()`](struct.SecretKey.html).
 /// - The length of the messages is leaked.
 /// - It is recommended to use `StreamTag::FINISH` as tag for the last message. This allows the
@@ -413,24 +416,37 @@ mod public {
 			assert_eq!(tag, StreamTag::MESSAGE);
 		}
 
-		/* TODO, decision about empty ciphertext
 		#[test]
-		fn test_auth_enc_plaintext_empty_err() {
+		fn test_seal_chunk_plaintext_empty_ok() {
 			let key = SecretKey::default();
-			let plaintext = "".as_bytes().to_vec();
+			let (mut sealer, _) = StreamSealer::new(&key).unwrap();
+			let plaintext = "".as_bytes();
 
-			assert!(seal(&key, &plaintext).is_err());
-		}*/
+			assert!(sealer.seal_chunk(plaintext, StreamTag::MESSAGE).is_ok());
+		}
 
-		//TODO, decision about empty ciphertext
 		#[test]
-		fn test_auth_enc_ciphertext_less_than_41_err() {
+		fn test_open_chunk_less_than_abytes_err() {
 			let key = SecretKey::default();
-			let ciphertext = [0u8; 16];
+			let ciphertext = [0u8; aead::streaming::ABYTES - 1];
 			let (_, nonce) = StreamSealer::new(&key).unwrap();
 			let mut opener = StreamOpener::new(&key, &nonce).unwrap();
 
 			assert!(opener.open_chunk(&ciphertext).is_err());
+		}
+
+		#[test]
+		fn test_open_chunk_abytes_exact_ok() {
+			let key = SecretKey::default();
+			let (mut sealer, nonce) = StreamSealer::new(&key).unwrap();
+			let mut opener = StreamOpener::new(&key, &nonce).unwrap();
+			let ciphertext = sealer
+				.seal_chunk("".as_bytes(), StreamTag::MESSAGE)
+				.unwrap();
+			let (pt, tag) = opener.open_chunk(&ciphertext).unwrap();
+
+			assert!(pt.is_empty());
+			assert!(tag.as_byte() == 0u8);
 		}
 
 		#[test]
--- a/src/aead.rs
+++ b/src/aead.rs
@@ -511,6 +511,25 @@ mod public {
 		}
 
 		#[test]
+		fn same_input_on_same_init_different_ct() {
+			// Two sealers initialized that encrypt the same plaintext
+			// should produce different ciphertexts because the nonce
+			// is randomly generated.
+			let key = SecretKey::default();
+			let (mut sealer_first, _) = StreamSealer::new(&key).unwrap();
+			let (mut sealer_second, _) = StreamSealer::new(&key).unwrap();
+			let plaintext = "Secret message 1".as_bytes();
+
+			let cipher1 = sealer_first
+				.seal_chunk(plaintext, StreamTag::MESSAGE)
+				.unwrap();
+			let cipher2 = sealer_second
+				.seal_chunk(plaintext, StreamTag::MESSAGE)
+				.unwrap();
+			assert!(cipher1 != cipher2);
+		}
+
+		#[test]
 		fn test_stream_seal_and_open() {
 			let key = SecretKey::default();
 			let (mut sealer, nonce) = StreamSealer::new(&key).unwrap();
--- a/src/aead.rs
+++ b/src/aead.rs
@@ -156,12 +156,11 @@ pub mod streaming {
 	//!
 	//! # Use case:
 	//!  This can be used to encrypt and authenticate a stream of data. It prevents the
-	//!  modification, reordering, dropping or duplicating of messages. Nonce management is handled automatically
+	//!  modification, reordering, dropping or duplication of messages. Nonce management is handled automatically.
 	//!
 	//!  An example of this could be the encryption of files which are too large to encrypt in one piece.
 	//!
 	//! # About:
-	//!
 	//! This implementation is based on and compatible with the ["secretstream" API](https://download.libsodium.org/doc/secret-key_cryptography/secretstream)
 	//! of libsodium.
 	//!
@@ -176,14 +175,14 @@ pub mod streaming {
 	//! # Errors:
 	//! An error will be returned if:
 	//! - `secret_key` is not 32 bytes.
-	//! - The length of `ciphertext` is not greater than `16`.
+	//! - The length of `ciphertext` is not at least [`ABYTES`].
 	//! - The received mac does not match the calculated mac when decrypting. This can indicate
 	//!   a dropped or reordered message within the stream.
 	//! - More than 2^32-3 * 64 bytes of data are processed when encrypting/decrypting a single chunk.
+	//! - [`ABYTES`] + `plaintext.len()` overflows when encrypting.
 	//!
 	//! # Panics:
 	//! A panic will occur if:
-	//! - [`ABYTES`] + `plaintext.len()` overflows when encrypting.
 	//! - 64 + (`ciphertext.len()` - [`ABYTES`]) overflows when decrypting.
 	//! - Failure to generate random bytes securely.
 	//!
--- a/src/aead.rs
+++ b/src/aead.rs
@@ -166,8 +166,6 @@ pub fn open(
 /// # Parameters:
 /// - `secret_key`: The secret key.
 /// - `nonce`: The nonce value.
-/// - `ad`: Additional data to authenticate (this is not encrypted and can be
-///   `None`. This data is also not a part of the output).
 /// - `plaintext`: The data to be encrypted.
 /// - `ciphertext`: The encrypted data with a Poly1305 tag and a [`StreamTag`](./hazardous/aead/streaming/enum.StreamTag.html) indicating its function.
 /// - `tag`: Indicates the type of message. The `tag` is a part of the output when encrypting. It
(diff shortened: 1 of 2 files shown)


src/aead.rs
```diff
@@ -148,13 +149,179 @@ pub fn open(
...
+	}
+
+	#[must_use = "SECURITY WARNING: Ignoring a Result can have real security implications."]
+	/// Initialize a `StreamSealer` struct with a given key.
+	pub fn new(secret_key: &SecretKey) -> Result<Self, UnknownCryptoError> {
+		let nonce = Nonce::generate();
+		let sk = &aead::streaming::SecretKey::from_slice(secret_key.unprotected_as_bytes())?;
+
+		Ok(Self {
+			internal_sealer: aead::streaming::StreamXChaCha20Poly1305::new(&sk, &nonce),
```
brycx: `&sk` -> `sk` (needless reference)

src/aead.rs
```diff
@@ -148,13 +149,179 @@ pub fn open(
...
+}
+
+impl StreamOpener {
+	#[must_use = "SECURITY WARNING: Ignoring a Result can have real security implications."]
+	/// Initialize a `StreamOpener` struct with a given key and nonce
+	pub fn new(secret_key: &SecretKey, nonce: &Nonce) -> Result<Self, UnknownCryptoError> {
+		let sk = &chacha20::SecretKey::from_slice(secret_key.unprotected_as_bytes())?;
+
+		Ok(Self {
+			internal_sealer: aead::streaming::StreamXChaCha20Poly1305::new(&sk, &nonce),
```
brycx: `&sk, &nonce` -> `sk, nonce` (needless references)

src/aead.rs
```diff
@@ -148,13 +149,179 @@ pub fn open(
...
+			.seal_chunk(plaintext, ad, &mut sealed_chunk, tag)?;
+
+		Ok(sealed_chunk)
+	}
+}
+
+/// TODO: Link does not work on index site
+/// Decrypts messages encrypted by a `StreamSealer`.
+/// See [`StreamSealer`](struct.StreamSealer.html) for documentation
+pub struct StreamOpener {
```
brycx: Please derive `Debug` implementation.

src/aead.rs
```diff
@@ -148,13 +149,179 @@ pub fn open(
...
+/// let cipher3 = sealer
+///		.seal_chunk(&plaintext3, None, StreamTag::FINISH)
+///		.unwrap();
+/// let (dec3, tag3) = opener.open_chunk(&cipher3, None).unwrap();
+/// assert_eq!(plaintext3, dec3);
+/// assert_eq!(tag3, StreamTag::FINISH);
+///
+/// # Ok::<(), orion::errors::UnknownCryptoError>(())
+/// ```
+pub struct StreamSealer {
```
brycx: Please derive `Debug` implementation.

src/aead.rs
```diff
@@ -148,13 +149,179 @@ pub fn open(
...
+		self.internal_sealer
+			.seal_chunk(plaintext, ad, &mut sealed_chunk, tag)?;
+
+		Ok(sealed_chunk)
+	}
+}
+
+/// TODO: Link does not work on index site
+/// Decrypts messages encrypted by a `StreamSealer`.
+/// See [`StreamSealer`](struct.StreamSealer.html) for documentation
```
brycx: This link works fine on the `aead` index site for me. Any specific issue you encountered?

brycx:
> I'm a bit unsure if it is better to have a get_nonce method or if it is better to return Self and the Nonce in the new method.

Thinking about it, it may be better to simply return a `(Self, Nonce)` tuple on `new`, since a user would always have to store the nonce somewhere in order to decrypt the stream later. Having a `get_nonce` method seems to be more appropriate when it is optional to store it. Which it isn't in this API, because it's randomly generated and the user has no control over that.

> Tests a rather minimal for now, but i think this should be fine since its only a thin wrapper around the hazardous API

This is true. Though I'd still like to test things such as errors, for example on invalid key-lengths. This is mostly to establish some baseline of the API, so that if we accidentally change some of the underlying components, the tests in this module would detect it (I mean avoiding accidental breaking changes to the public API). You can add such tests if you want, otherwise I'll take care of it before merging.

> Is it useful to expose the option to authenticate additional data? The "normal" AEAD seal function does not do it

I'm glad you caught this, because this would indeed bring some inconsistency to the API of this module. Let's remove the `ad` parameter.

snsmac:
I have one more question: Currently the AEAD function and the new stream function panic if the input is too large. Why don't we check the size of the input and return an error instead?

brycx:
Sorry for getting back to you this late.

> Why don't we check the size of the input and return an error instead?

I actually don't know why we're panicking here instead of returning an error. Seeing as these panicking operations are on `usize` and we support 32-bit (where `usize == u32` if I'm not mistaken), we actually panic in AEAD on inputs that don't fit ~0.5 GB (size of `u32::max_value()`) IIUC. I still think it's fine to panic in `generate_auth_tag()` due to the cast to `u64`, such that the overflowing limit is ~2305843009 GB. Therefor it seems that these checks need to return an error on overflow instead of panic:

https://github.com/brycx/orion/blob/3cd6c1ff13e9d70f493d76ccef906354f9002c5e/src/hazardous/aead/streaming.rs#L306

https://github.com/brycx/orion/blob/3cd6c1ff13e9d70f493d76ccef906354f9002c5e/src/hazardous/aead/chacha20poly1305.rs#L182

You may include this change in this PR, otherwise it can just be a separate commit.

brycx:
> /// - It is recommended to use `StreamTag::FINISH` as tag for the last message. This allows the
///   decrypting side to detect if messages at the end of the stream are lost.

This made me think about whether we could force the user to close a stream with a finishing tag.

What I've come up with so far is having a `seal_final_chunk()` method that returns the nonce used for the stream, instead of the `new()` method. This way a user has to use the final method to be able to decrypt the stream later on and the chunks at the end cannot be dropped.

Then having the StreamSealer/StreamOpener structs track whether they have been finalised, using a bool.

The StreamOpener struct would have only `open_chunk()` and return an `Option` in the `Result` it returns now, giving `None` if the stream has been finalised already, and `Some(Vec<u8>)` if not. Though I fear returning `Result<Option<Vec<u8>, UnknownCryptoError>` will require too much matching on the output and discourage users from using this API and instead turning to the hazardous.

What do you think about this?

snsmac:
> What I've come up with so far is having a `seal_final_chunk()` method that returns the nonce used for the stream, instead of the `new()` method. This way a user has to use the final method to be able to decrypt the stream later on and the chunks at the end cannot be dropped.

I don't think this is a good idea. For example if you want to send a large file (to large to encrypt all at once) over a network you can encrypt it in chunks and send them separately.  The reciver can start decrypting and processing this chunks before all chunks are transfered. With the seal_final_chunk() method this would not be possible anymore

brycx:
It's true that this specific use case would be impossible with the forced finish tag. Do you know how common this use case is? If it's rather niche then maybe it's not worth sacrificing misuse-resistance and instead opt to refer users to the hazardous implementation, for that specifically.

snsmac:
@brycx Generally this change would prevent any usecase that wants to start processing data before all encrypted chunks are available which i think is a rather common. But i think it could be useful to add an example which shows how to decrypt data in a loop until a finish tag is recived

brycx:
I've tried searching a bit, but haven't been able to find any information on how common the use case is. Let's keep it as-is, also making it compatible with libsodium's API.

> But i think it could be useful to add an example which shows how to decrypt data in a loop until a finish tag is recived

I agree. Could you change the current example to reflect this? I'll add this to the high-level API fuzzers and look into adding some more tests, if applicable. Once I've done that, this should be ready for merge.

src/aead.rs
```diff
@@ -148,13 +149,170 @@ pub fn open(
...
+/// # Panics:
+/// A panic will occur if:
+/// - [`ABYTES`](../hazardous/aead/streaming/constant.ABYTES.html) + `plaintext.len()` overflows when encrypting.
+/// - 64 + (`ciphertext.len()` - [`ABYTES`](../hazardous/aead/streaming/constant.ABYTES.html)) overflows when decrypting.
+/// - Failure to generate random bytes securely.
+///
+/// # Security:
+/// - To securely generate a strong key, use [`SecretKey::generate()`](struct.SecretKey.html).
+/// - The length of the messages is leaked.
+/// - It is recommended to use `StreamTag::FINISH` as tag for the last message. This allows the
```
brycx: Let's add this important point to the hazardous API docs as well.

src/aead.rs
```diff
@@ -148,13 +149,170 @@ pub fn open(
...
+/// - `ciphertext`: The encrypted data with a Poly1305 tag and a [`StreamTag`](./hazardous/aead/streaming/enum.StreamTag.html) indicating its function.
+/// - `tag`: Indicates the type of message. The `tag` is a part of the output when encrypting. It
+/// is encrypted and authenticated.
+///
+/// # Errors:
+/// An error will be returned if:
+/// - The length of `ciphertext` is not greater than `16`.
+/// - The received mac does not match the calculated mac when decrypting. This can indicate
+///   a dropped or reordered message within the stream.
+/// - More than 2^32-3 * 64 bytes of data are processed when encrypting/decrypting a single chunk.
```
brycx: We should also document that an error is returned if `SecretKey` is not 32 bytes in length.

brycx:
Let's leave the fuzzing for this target for now. I can't think of any target that would give any value to fuzzing, in addition to the hazardous fuzzing target.

src/aead.rs
```diff
@@ -219,30 +377,204 @@ mod public {
...
+			let dst_ciphertext = sealer.seal_chunk(plaintext, StreamTag::MESSAGE).unwrap();
+			assert!(dst_ciphertext.len() == plaintext.len() + 17);
+			let (dst_plaintext, tag) = opener.open_chunk(&dst_ciphertext).unwrap();
+			assert_eq!(plaintext, &dst_plaintext[..]);
+			assert_eq!(tag, StreamTag::MESSAGE);
+		}
+
+		/* TODO, decision about empty ciphertext
+		#[test]
+		fn test_auth_enc_plaintext_empty_err() {
```
brycx: IMO we should leave it as-is. Supporting empty ciphertexts, but still require the `Tag`. This way we support use cases where someone may need to pass information without any real encrypted data, such as rekey tags during sessions. I believe this is also what you suggested to begin with.

brycx:
I haven't been able to make an example for the docs, regarding the check for `StreamTag::FINISH`, that I was satisfied with. It seems whatever I come up with ends up cluttering the examples too much. Let's keep the current one that you included @snsmac, then we can take a look at making the other one at a later point.

All other comments should be addressed now and the PR should be ready for merge.

PR #110, opened 2019-11-18
AEAD: Err instead of panic on input length addition overflow

brycx:
Discussed in https://github.com/brycx/orion/pull/108.


brycx:
@snsmac Could you take a look at this? I want to make sure this is how we both imagined it.

PR #113, opened 2020-01-04
Argon2i

### LINKED ISSUE #109: Add support for new password hashing algorithm
Currently the only password hashing algorithm supported is PBKDF2. This is a dated algorithm and better alternatives exist.

The high-level API in `orion::pwhash` offers key derivation using PBKDF2, but should preferably use a more modern algorithm.

Argon2i would be a very good replacement, which is also provided by other crypto libraries such as libsodium. I know @vlmutolo began some work on implementing this. 

To provide Argon2i and keep unsafe code at a minimum, this should for starters not support threading. Alternatively we could provide scrypt, though a PBKDF2-SHA256 implementation is missing for this. I don't see an argument for providing scrypt rather than Argon2i however.

brycx:
Closes #109.

TODO:

- [X] Argon2i single-threaded in `hazardous`. How should the allocations for the memory blocks be handled?
- [X] Switch to Argon2i in high-level `orion::pwhash` and `orion::kdf`. What parameters should be used as default for each of those high-level APIs?
- [X] Fuzzers in orion-fuzz. 
- [X] Generated test vectors.

Edit: I haven't been able to get a solution working, that doesn't require heap allocation. Thus, the Argon2i implementation will not be available with `no_std`.


src/typedefs.rs
```diff
@@ -750,7 +750,7 @@ macro_rules! construct_secret_key_variable_size {
         /// # }
         /// ```
         pub struct $name {
-            value: Vec<u8>,
+            pub(crate) value: Vec<u8>,
```
brycx: This is set to `pub(crate)` so we avoid allocation two different `Vec<u8>`s in `orion::kdf` which hold sensitive data.

PR #137, opened 2020-12-03
Use encoded parameters in hash_password_verify

vlmutolo:
The `pwhash::hash_password_verify` function is central to the UX of the
password verification portion of this library. Currently, it requires callers
to specify the iterations and memory to be used in the Argon2i hash that the
function performs.

However, one of the arguments to the function is a `PasswordHash` type. This
type represents a password that has already been hashed, and it contains the
memory and iterations that were used to hash it. It's almost certainly
a logical error to hash a password for the purpose of comparing it to another
hash generated using different parameters. We can save users from having to
specify the paramters---and eventually making a mistake doing so---by having
`hash_password_verify` just pull the parameters from the given `PasswordHash`.
We retain the original functionality in a separate function called
`hash_password_verify_with`.

### DIFF
--- a/src/high_level/pwhash.rs
+++ b/src/high_level/pwhash.rs
@@ -425,50 +425,6 @@ pub fn hash_password_verify(
     )
 }
 
-/// Hash and verify a password using Argon2i. If you want the parameters for the Argon2i
-/// hash to be automatically detected (which you likely do), please see the
-/// [hash_password_verify](crate::pwhash::hash_password_verify) function.
-///
-/// # Example:
-/// ```rust
-/// use orion::pwhash;
-///
-/// let password = pwhash::Password::from_slice(b"Secret password")?;
-/// let hash = pwhash::hash_password(&password, 3, 1<<16)?;
-///
-/// // Correct paramters
-/// assert!(pwhash::hash_password_verify_with(&hash, &password, 3, 1<<16).is_ok());
-///
-/// // Incorrect parameters
-/// assert!(pwhash::hash_password_verify_with(&hash, &password, 3, 1<<15).is_err());
-/// assert!(pwhash::hash_password_verify_with(&hash, &password, 4, 1<<16).is_err());
-/// # Ok::<(), orion::errors::UnknownCryptoError>(())
-/// ```
-#[must_use = "SECURITY WARNING: Ignoring a Result can have real security implications."]
-pub fn hash_password_verify_with(
-    expected: &PasswordHash,
-    password: &Password,
-    iterations: u32,
-    memory: u32,
-) -> Result<(), UnknownCryptoError> {
-    if iterations < MIN_ITERATIONS {
-        return Err(UnknownCryptoError);
-    }
-
-    let mut buffer = Zeroizing::new([0u8; PWHASH_LENGTH]);
-
-    argon2i::verify(
-        expected.unprotected_as_bytes(),
-        password.unprotected_as_bytes(),
-        expected.salt.as_ref(),
-        iterations,
-        memory,
-        None,
-        None,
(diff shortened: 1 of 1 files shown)


vlmutolo:
If this isn't a direction you're interesting in heading, please just ignore the PR and/or close it. I saw that the change would be pretty simple to implement it, and figured it would be easier to discuss with a working example.

vlmutolo:
Resolves #138

src/high_level/pwhash.rs
```diff
@@ -844,16 +894,21 @@ mod public {
             salt_mod[0..16].copy_from_slice(&[0u8; 16]);
             let modified = PasswordHash::from_slice(&pwd_mod, &salt_mod, 3, 4096).unwrap();
 
-            assert!(hash_password_verify(&modified, &password, 3, 4096).is_err());
+            assert!(hash_password_verify(&modified, &password).is_err());
```
brycx, line 897: I think it would be good to add a test, similar to the suggestion in the doc example, that shows modifying the `iterations` and `memory` parameters (individually and at the same time), also gives a verification error.

brycx (changes requested):
Thanks so much for the PR @vlmutolo!

Apart from the few small NITs, it would be nice if we could merge this into `v016` instead of `master`.

src/high_level/pwhash.rs
```diff
@@ -382,11 +382,71 @@ pub fn hash_password(
     PasswordHash::from_slice(buffer.as_ref(), salt.as_ref(), iterations, memory)
 }
 
+/// Hash and verify a password using Argon2i. The Argon2i parameters "iterations"
+/// and "memory" will be pulled from the `expected: &PasswordHash` argument. If
+/// you want to manually specify the iterations and memory for Argon2i to use in
+/// hashing the `password` argument, see the
+/// [hash_password_verify_with](crate::pwhash::hash_password_verify_with) function.
```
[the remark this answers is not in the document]
vlmutolo: I was so excited when they announced this feature.

src/high_level/pwhash.rs
vlmutolo: I'm not sure how to do this without using the `hazardous` module. Is that what you're suggesting?

brycx: I was thinking more in the lines of taking one of the existing, valid, encoded password hashes and manually modifying the iterations/memory parameters therein (modifying within in the string). Like one of the encoded strings you've removed here as a starting point perhaps?

brycx:
There seem to be some issues regarding clippy. I'll take care of those on `master` and rebase `v016` thereon.

brycx:
@vlmutolo I've fixed the clippy warnings and added an explanation to my last CR-comment. If you could rebase this branch on the changes now in `v016` and add the missing test, then this should be ready to merge. The rest looks very good!

src/high_level/pwhash.rs
vlmutolo: Modifying the encoded parameters turned out to be a good way to do it. Nice and simple. Good idea

brycx (approved):
Looks awesome! I'll re-introduce the tests for `hash_password()` that check errors on `MIN_ITERATIONS - 1` and `MIN_MEMORY - 1`, because I didn't see any of the other tests checking this.


Note to make sure this change i tracked:

Before this change, `hash_password_verify()` checked the `iterations` (then, a separate param):

```
if iterations < MIN_ITERATIONS {
        return Err(UnknownCryptoError);
}
```

This check is now omitted because `iterations` is no longer a separate parameter, and `PasswordHash` already checks if the iterations count is at least `MIN_ITERATIONS` when created.

PR #151, opened 2020-12-09
Remove the badges from Cargo #148

### LINKED ISSUE #148: Remove badges from Cargo.toml
According to the [cargo docs](https://doc.rust-lang.org/cargo/reference/manifest.html#the-badges-section):

> Note: crates.io previously displayed badges next to a crate on its website, but that functionality has been removed. Packages should place badges in its README file which will be displayed on crates.io (see the readme field).

So the `[bagdes]` field from `Cargo.toml` needs to be deleted. The badges therein also point to a no-longer existing branch.

u5surf:
Fixes #148 

I removed the field of badges in Cargo.toml.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,8 +41,3 @@ harness = false
 [profile.dev]
 opt-level = 1
 
-[badges]
-travis-ci = { repository = "brycx/orion", branch = "crates-published" }
-codecov = { repository = "brycx/orion", branch = "crates-published" }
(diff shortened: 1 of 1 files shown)


brycx (approved):
Thanks for this @u5surf!

I'm glad to see you're picking up more issues. Please let me know if there are some others you're interested working on, but seem a bit confusing. I'm happy to mentor if I can.

PR #160, opened 2020-12-19
Replace T::max_value() with MAX (#156)

### LINKED ISSUE #156: Replace uses of T::max_value() with T::MAX
The max value that a primitive type can represent has been made [accessible](https://doc.rust-lang.org/std/primitive.u32.html#associatedconstant.MAX) as an associated constant. The `max_value()`/`min_value()` functions are soft-deprecated.

```rust
u32::max_value()
```
should change to
```rust
u32::MAX
```

If `T::min_value()` is used anywhere, this should be replaced with `T::MIN`.

Bumps MSRV to 1.43.0.

u5surf:
Fixes #156

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,7 +21,7 @@ jobs:
           - stable
           - beta
           - nightly
-          - 1.41.0 # MSRV
+          - 1.43.0 # MSRV
         os: 
           - ubuntu-latest
           - macos-latest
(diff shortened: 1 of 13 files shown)


brycx (commented):
Thanks for doing this @u5surf! Please also see my comment in the original issue, about modifying the MSRV for CI.

Additionally, it would be nice to update the README section and the badge therein, as well, to reflect the bump in MSRV.

brycx (approved):
Looks good @u5surf, thanks!

PR #173, opened 2021-02-03
Doc: Change the example in orion::auth

### LINKED ISSUE #159: Change the example in orion::auth
The example for `orion::auth` should be updated to more clearly express the origin of the expected tag, used when verifying. See [this comment for information](https://github.com/brycx/orion/issues/157#issuecomment-746326029).

@vlmutolo presented the following change and will also work on implementing this:
```rust
use orion::auth;

// There exists a shared key
let key = auth::SecretKey::default();

// Unauthenticated source generates message and authentication tag
let msg = "Some message.".as_bytes();
let expected_tag = auth::authenticate(&key, msg)?;

// Verify the authenticity of the tag
assert!(auth::authenticate_verify(&expected_tag, &key, &msg).is_ok());
```

brycx:
Closes #159 

@vlmutolo Does this change make sense, is it clear? I've extended upon your suggestion a bit, in order to make the comments align more the the use-case description/example in the top of the module docs.

> `orion::auth` can be used to ensure message integrity and authenticity by
using a secret key.
>
> An example of this could be securing APIs by having a user of a given API
sign their API request and having the API server verify these signed API
requests.

### DIFF
--- a/src/high_level/auth.rs
+++ b/src/high_level/auth.rs
@@ -57,14 +57,14 @@
 //! ```rust
 //! use orion::auth;
 //!
-//! // There exists a shared key between the user and API endpoint
+//! // There exists a shared key between the user and API server
 //! let key = auth::SecretKey::default();
 //!
 //! // User generates message and authentication tag
 //! let msg = "Some message.".as_bytes();
 //! let expected_tag = auth::authenticate(&key, msg)?;
 //!
-//! // API endpoint verifies the authenticity of the message with the tag
+//! // API server verifies the authenticity of the message with the tag
 //! assert!(auth::authenticate_verify(&expected_tag, &key, &msg).is_ok());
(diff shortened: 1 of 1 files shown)


vlmutolo:
Yes, that makes sense to me. I think it’s a good idea to specify the two parties sharing the key, like you did. I wonder if something like “caller of API” is clearer than “API endpoint”. The word “endpoint” is throwing me a little.

brycx:
> I wonder if something like “caller of API” is clearer than “API endpoint”. The word “endpoint” is throwing me a little.

I was split by this one too. "caller" seemed "less correct", since the user usually makes calls/request to the API. But "API server" seems weird too.

brycx:
I've looked around some more and only really found reference to the API in form of "server". For example, [this post](https://www.oauth.com/oauth2-servers/definitions/) for OAuth calling it "resource server". I'll change it to "server" and use that.

PR #186, opened 2021-03-15
Documentation for correct use of streaming AEAD API with StreamTag::Finish

### LINKED ISSUE #139: Adding documentation for correct use of streaming AEAD API with StreamTag::FINISH
This was originally a suggestion by @snsmac, which I somehow lost in a backlog (sorry about that!).

Using the streaming AEAD API securely for encryption, relies on that fact that the end of a stream is marked correctly with `StreamTag::FINISH` to avoid truncation issues. This is mentioned in the security section of both the `hazardous` and high-level APIs. But @snsmac thought that it would be good to extend the usage examples with something that encrypts a stream in a loop, adds the `StreamTag::FINISH`, and then decrypts this with a check for that same tag.

Since Orion implements the libsodium secretstream, the example could look similar to [that of libsodium](https://doc.libsodium.org/secret-key_cryptography/secretstream#file-encryption-example-code).

brycx:
closes #139 

#139 mentions the current example should be extended, but I felt like it grew so long it might actually have users turn away from it.

cc @vlmutolo, @snsmac

### DIFF
--- a/src/high_level/aead.rs
+++ b/src/high_level/aead.rs
@@ -204,11 +204,11 @@ pub mod streaming {
     //!     let encrypted_chunk =
     //!         if src_chunk.len() != chunk_size || n_chunk + 1 == src.len() / chunk_size {
     //!             // We've reached the end of the input source,
-    //!             // so we mark it with the FINISH tag.
-    //!             sealer.seal_chunk(src_chunk, StreamTag::FINISH).unwrap()
+    //!             // so we mark it with the Finish tag.
+    //!             sealer.seal_chunk(src_chunk, StreamTag::Finish).unwrap()
     //!         } else {
     //!             // Just a normal chunk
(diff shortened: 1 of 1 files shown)


brycx:
This PR has been sitting around for a while now. Since no comments have been added to it, I've updated what was missing and will merge it now.

vlmutolo:
For what it's worth, I just looked at the diff here and I think the docs are good. It's definitely long, but I was able to read through it and understand what was going on with very little trouble.

brycx:
> For what it's worth, I just looked at the diff here and I think the docs are good. It's definitely long, but I was able to read through it and understand what was going on with very little trouble.

Thanks for taking a look. I agree it's a bit long, but I couldn't find a slimmer yet clear way to express it. Let me know if you come up with anything by chance.

PR #187, opened 2021-03-15
KDF output buffer zeroization documentation

### LINKED ISSUE #178: Document that out-parameters used with verification in KDFs should be zeroized
The `verify()` functions offered in `hazardous::kdf` require an out-parameter in order to support `no_std`. During verification, some amount of bytes are derived from a secret, which are placed into the out-parameter. These are then used to compare with an expected value.

Using the `derive_key()` functions, it is obvious that the out-parameter will contain the secret key, but it may not be obvious when using `verify()`. Therefor, the security section of documentation for all `hazardous::kdf` modules should mention that, sensitive values are placed into the out-parameter, but Orion does not zeroize them. So users must make sure they do if needed.

Example:
```
[..]
- When `verify()`ing, the `dst_out` parameter should always be zeroed out after finishing the operation.
```

However, for similar things, Orion usually offers utility functions such as `secure_rand_bytes` or `secure_cmp`. The question is whether Orion should offer a zeroization method (wrapper around `zeroize` which Orion uses internally) or simply point users to the [`zeroize` crate](https://crates.io/crates/zeroize)?

brycx:
closes #178 

cc @vlmutolo

### DIFF
--- a/src/hazardous/kdf/argon2i.rs
+++ b/src/hazardous/kdf/argon2i.rs
@@ -60,7 +60,7 @@
 //! - Password hashes should always be compared in constant-time.
 //! - Please note that when verifying, a copy of the computed password hash is placed into
 //! `dst_out`. If the derived hash is considered sensitive in the given context, the user is
-//! responsible for zeroing out this buffer (see [`zeroize crate`]).
+//! responsible for zeroing out this buffer (see [`zeroize` crate]).
 //!
 //! # Example:
 //! ```rust
@@ -89,7 +89,7 @@
 //! # Ok::<(), orion::errors::UnknownCryptoError>(())
 //! ```
 //! [`secure_rand_bytes()`]: crate::util::secure_rand_bytes
-//! [`zeroize crate`]: https://crates.io/crates/zeroize
+//! [`zeroize` crate]: https://crates.io/crates/zeroize
 
 use crate::errors::UnknownCryptoError;
 use crate::hazardous::hash::blake2b::{Blake2b, BLAKE2B_OUTSIZE};
--- a/src/hazardous/kdf/argon2i.rs
+++ b/src/hazardous/kdf/argon2i.rs
@@ -58,6 +58,9 @@
 //! - The minimum recommended length for a hashed password is `16` bytes.
 //! - The minimum recommended iteration count is `3`.
 //! - Password hashes should always be compared in constant-time.
+//! - Please note that when verifying, a copy of the computed password hash is placed into
+//! `dst_out`. If the derived hash is considered sensitive in the given context, the user is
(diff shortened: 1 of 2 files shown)


src/hazardous/kdf/argon2i.rs
```diff
@@ -58,6 +58,9 @@
 //! - The minimum recommended length for a hashed password is `16` bytes.
 //! - The minimum recommended iteration count is `3`.
 //! - Password hashes should always be compared in constant-time.
+//! - Please note that when verifying, a copy of the computed password hash is placed into
+//! `dst_out`. If the derived hash is considered sensitive in the given context, the user is
+//! responsible for zeroing out this buffer (see [`zeroize crate`]).
```
vlmutolo, line 63: Small formatting nit: I think `[\`zeroize crate\`]` should be `[\`zeroize\` crate]`

vlmutolo: ## Proposal
Consider changing the message slightly to:
> If the derived hash is considered sensitive and you want to provide defense in depth against an attacker reading your application's private memory, the you as the user are responsible for zeroing out this buffer (see [`zeroize crate`]).

## Justification

There are two necessary conditions for an orion user to care about the information in the `dst_out` buffer.

1. The information has to be considered sensitive.
2. The `dst_out` buffer has to be considered vulnerable to compromise.

For user password-hashing applications, it's reasonable to consider the information sensitive. Even though this is for a secure password hash, the best hash is the one that an attacker doesn't have. Also, though it's unlikely that an attacker will gain access to the private memory of the application and be able to actually read the `dst_out` buffer, it does happen in practice. Other vulnerabilities in the same process can lead to the compromise of that buffer.

I think it's a good idea to be specific in what exactly the user would be protecting against by going off to use the `zeroize` crate.

vlmutolo: Also, I considered suggesting that there should be a note that says something like "You likely don't have to use this." I decided against that suggestion, though, when I considered that this is in the `hazardous` module, and it's fair to assume that the people using it know what they're doing.

brycx: > I think it's a good idea to be specific about what exactly the user would be protecting against by going off to use the zeroize crate.

Very nice addition to the documentation. It really does make it much clearer and I feel like I'm able to much more quickly ascertain whether or not this is something I _do_ care about. Actually makes a world of difference.

brycx: > Also, I considered suggesting that there should be a note that says something like "You likely don't have to use this." I decided against that suggestion, though, when I considered that this is in the hazardous module, and it's fair to assume that the people using it know what they're doing.

Like i mentioned in #178, this part I had forgotten about. I completely agree with your reasoning.

PR #189, opened 2021-03-23
Use constant-time Base64 in `pwhash::PasswordHash`

### LINKED ISSUE #188: Constant-time Base64-encoding for Argon2i password hash
The current encoding and decoding of a `PasswordHash` in `orion::pwhash` is not constant-time (also noted in the security section of the docs). It's been in the backlog to implement this, but it seems a viable crate has been released in the meantime: https://github.com/jedisct1/rust-ct-codecs

According to the README, the crate re-implements the encoding formats from libsodium/libhydrogen. It also defines`#![forbid(unsafe_code)]`.

TODO:
- The current `base64` crate should be switched out with the above
- The wikis "Dependencies" section should also be updated to reflect this change
- Add tests in https://github.com/brycx/orion-dudect for constant-time execution

brycx:
Replace the current Base64 dependency with one that provides constant-time encoding & decoding.

closes #188

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -53,8 +53,8 @@ impl From<getrandom::Error> for UnknownCryptoError {
 }
 
 #[cfg(feature = "safe_api")]
-impl From<base64::DecodeError> for UnknownCryptoError {
-    fn from(_: base64::DecodeError) -> Self {
+impl From<ct_codecs::Error> for UnknownCryptoError {
+    fn from(_: ct_codecs::Error) -> Self {
         UnknownCryptoError
     }
 }
@@ -104,11 +104,10 @@ fn test_source() {
 #[test]
 #[cfg(feature = "safe_api")]
 fn test_unknown_crypto_from_decode_error() {
-    use base64::DecodeError;
+    use ct_codecs::Error;
 
-    let err_one = DecodeError::InvalidByte(0, 0);
-    let err_two = DecodeError::InvalidLastSymbol(0, 0);
-    let err_three = DecodeError::InvalidLength;
+    let err_one = Error::InvalidInput;
+    let err_two = Error::Overflow;
 
     // Tests Debug impl through "{:?}" and Display impl though "{}"
     let err = format!(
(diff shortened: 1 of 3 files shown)


src/errors.rs
```diff
@@ -104,11 +104,10 @@ fn test_source() {
...
-    use base64::DecodeError;
+    use ct_codecs::Error;
 
-    let err_one = DecodeError::InvalidByte(0, 0);
-    let err_two = DecodeError::InvalidLastSymbol(0, 0);
-    let err_three = DecodeError::InvalidLength;
+    let err_one = Error::InvalidInput;
+    let err_two = Error::Overflow;
 
     // Tests Debug impl through "{:?}" and Display impl though "{}"
```
vlmutolo, line 112: This comment confused me. Are we still testing both `Debug` and `Display`, or is it just `Display` now?

src/errors.rs
```diff
@@ -123,12 +122,6 @@ fn test_unknown_crypto_from_decode_error() {
         UnknownCryptoError::from(err_two)
     );
     assert_eq!(err, "UnknownCryptoErrorUnknownCryptoError");
```
vlmutolo, line 124: Just out of curiosity, why is `UnknownCryptoError` repeated twice?

vlmutolo (approved):
Looks good to me! The only real suggestion I had was to take a look at a comment that I think was invalidated by the changes in this PR. Definitely non-critical, though.

src/errors.rs
brycx: We're still testing both. It's just that in the later `format!` call, the note explains which formatter is used to test which impl, if that's makes sense.

src/errors.rs
brycx: In continuation of the previous comment, the two formatters are `{:?}{}`, which means `"UknownCryptoError" is "printed two times in the string", once using the `Debug` impl, the other using the `Display` impl. I'll separate these with a `":"`, perhaps that would make it clearer.

brycx: Thanks for reviewing @vlmutolo, always a big help. I'll add the separator mentioned above. Let me know if it's not clearer or something else would be better. I'll merge it with that for now though, so I can get this included for the fuzzing.

src/errors.rs
vlmutolo: Oops! For some reason, I thought that the format string had changed. I see it now.

PR #193, opened 2021-03-27
add initial serde support for relevant types

### LINKED ISSUE #192: Implement serde Serialize and Deserialize for relevant types
## Summary

The `serde::Serialize` and `serde::Deserialize` traits are used pervasively throughout the Rust ecosystem to enable flexible (de)serialization of data. Many authors derive these types on structs using macros provided by `serde`. If orion's types don't support these traits, users won't be able to derive these serialization traits, and will have to resort to converting orion types before storing them in a struct for serialization.

By implementing `serde`'s standard traits, we enable authors to skip the mental burden and increased complexity of serializing orion types without serde.

## Scope
At least for a first implementation, I would suggest only implementing serialization traits for types that are expected to either be stored in a database or sent to another application (e.g. over the network). This would include the following types.
- `pwhash::PasswordHash`
- `hash::Digest`
- `auth::Tag`
- `kdf::Salt` (added after suggestion below)

## Motivating Example
My motivating example is storing a `User` type in a database. It would be nice to simply define a `User` struct, and then have it serialize to one long string using, for example, the `bincode` crate. This is what I would like the basic code to look like (using a fictional database API).

```rust
use serde::{Serialize, Deserialize};
use orion::pwhash::PasswordHash;

#[derive(Serialize, Deserialize)]
struct User {
    username: String,
    pwhash: PasswordHash,
    last_login: chrono::DateTime<Utc>,
}

fn store_user_in_db(user: &User, db: &Db) {
    db.write_bytes(&user.username, bincode::serialize(user));
}

fn retrieve_user_from_db(db: &Db) -> User {
    let bytes = db.get_bytes(&user.username);
    bincode::deserialize(bytes)
}
```

The `PasswordHash` type would serialize as a string by delegating to `PasswordHash::unprotected_as_encoded`.

Without `serde`'s types implemented, however, we would have to do one of the following.
1. Create a separate `User` type that can be serialized.
2. Implement `serde` manually on `User` and delegate the `PasswordHash` serialization to its `unprotected_as_encoded` method. Neither solution is particularly attractive.

A separate, serializable `User` struct is probably the easiest of the the alternatives. It would look like this.
```rust
// previous definitions omitted

#[derive(Serialize, Deserialize)]
struct UserSerialize {
    username: String,
    pwhash: String,
    last_login: chrono::DateTime<Utc>,
}

fn store_user_in_db(user: &User {username, pwhash, last_login}, db: &Db) {
    let user_serialize = UserSerialize {username, last_login, pwhash: pwhash.unprotected_as_encoded()};
    db.write_bytes(&user.username, bincode::serialize(to_serialize));
}

fn retrieve_user_from_db(db: &Db) -> User {
    let bytes = db.get_bytes(&user.username);
    let UserSerialize {username, pwhash, last_login} = bincode::deserialize(bytes);
    User {username, last_login, pwhash: PasswordHash::from_encoded(pwhash)}
}
```

It isn't terrible in terms of boilerplate, but it could start to get annoying and possibly error-prone to repeat struct definitions for several different types instead of just `User`.

Implementing [`Serialize`](https://serde.rs/impl-serialize.html) and [`Deserialize`](https://serde.rs/impl-deserialize.html) manually can be… unpleasant. Especially for anything complex, and definitely more boilerplate in any case.

## Pros of implementing `Serialize`/`Deserialize`
- There is significant ergonomic benefit to supporting this nearly ubiquitous pair of serialization traits.
- Users may be less likely to mistakenly reach for non-constant-time operations if they never actually (de)serialize the type themselves and just let `serde` do it for them. The common workflow is to deserialize some bytes or a string into the relevant type (would be `User` here, for example), and then to do operations on the fully featured type. This is beneficial because many of orion's types use constant-time operations for things like comparisons. It's anecdotal, but in my first example, the `PasswordHash` type spends very little time in its serialized state because it's so easy to deserialize into the fully featured type. Making deserialization the fast path could save users from security mistakes.

## Cons
- The serialization functions in orion are helpfully named `unprotected_*` to signify that by using them, authors are relinquishing the constant-time features that orion provides. If we hide these functions in a `serde::Serialize` implementation, users may be more likely to mistakenly serialize a sensitive type and perform insecure operations on it. I realize that this somewhat contradicts the last "Pro" above, but they both seem like legitimate possibilities to consider.

## Additional Considerations
- Serialization should conform to constant-time operations already in place. For example, `PasswordHash` should serialize as a string using the already-implemented `PasswordHash::unprotected_as_encoded` function.
- We may have to add additional mechanisms for testing to ensure that all tests pass with or without the hypothetical `serde` feature enabled.

vlmutolo:
Resolves #192.

This is still a draft PR. The following items should be addressed before merging.

[ ] No tests are currently implemented. I still have to think about what the best test cases would be. Suggestions are welcome.
[ ] When it was available for the given type, serialization implementations use `unprotected_as_bytes` over the `AsRef<[u8]>` trait impl. Should we prefer using `AsRef` for consistency with types that don't have `unproected_as_bytes`, or is there some reason to prefer `unprotected_as_bytes` for its semantics? I'm leaning toward just using `AsRef` for everything that converts to/from byte slices.
[ ] All the serde-related logic is implemented here in a separate module. I've taken to doing it this way in my own projects to keep the boilerplate out of the modules for each type, but I'm happy to break it up and move the four serde implementations nearer to where the types themselves are defined.
[ ] Another style point: I prefixed most of the types with their module names instead of importing them directly. For example, `kdf::Salt` instead of just `Salt`. I was concerned that other types may implement `Salt` in the future and we may add them to the `serde` module. This is a downside of keeping everything in one module; the names may have to be differentiated further.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ subtle = { version = "^2.2.2", default-features = false }
 zeroize = { version = "1.1.0", default-features = false }
 getrandom = { version = "0.2.0", optional = true }
 ct-codecs = { version = "1.1.1", optional = true }
+serde = { version = "1.0.124", optional = true }
 
 [features]
 default = [ "safe_api" ]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -74,6 +74,9 @@ extern crate quickcheck_macros;
 #[cfg_attr(feature = "alloc", macro_use)]
 extern crate alloc;
 
+#[cfg(feature = "serde")]
+mod serde_impls;
+
 #[macro_use]
 mod typedefs;
 
--- a/src/serde_impls.rs
+++ b/src/serde_impls.rs
@@ -1,0 +1,85 @@
+use crate::{auth, hash, kdf, pwhash::PasswordHash};
+use serde::{
+    de::{self, Deserialize, Deserializer},
+    ser::{Serialize, Serializer},
+};
+
+impl Serialize for PasswordHash {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let encoded_string = self.unprotected_as_encoded();
+        serializer.serialize_str(&encoded_string)
+    }
+}
+
+impl<'de> Deserialize<'de> for PasswordHash {
+    fn deserialize<D>(deserializer: D) -> Result<PasswordHash, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let encoded_str = <&str>::deserialize(deserializer)?;
+        PasswordHash::from_encoded(encoded_str).map_err(de::Error::custom)
+    }
+}
+
+impl Serialize for hash::Digest {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let bytes: &[u8] = self.as_ref();
+        bytes.serialize(serializer)
+    }
+}
+
+impl<'de> Deserialize<'de> for hash::Digest {
+    fn deserialize<D>(deserializer: D) -> Result<hash::Digest, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let bytes = <&[u8]>::deserialize(deserializer)?;
+        hash::Digest::from_slice(bytes).map_err(de::Error::custom)
+    }
+}
+
+impl Serialize for auth::Tag {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let bytes: &[u8] = self.unprotected_as_bytes();
+        bytes.serialize(serializer)
+    }
+}
+
+impl<'de> Deserialize<'de> for auth::Tag {
+    fn deserialize<D>(deserializer: D) -> Result<auth::Tag, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let bytes = <&[u8]>::deserialize(deserializer)?;
+        auth::Tag::from_slice(bytes).map_err(de::Error::custom)
+    }
+}
+
+impl Serialize for kdf::Salt {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let bytes: &[u8] = self.as_ref();
+        bytes.serialize(serializer)
+    }
+}
+
+impl<'de> Deserialize<'de> for kdf::Salt {
+    fn deserialize<D>(deserializer: D) -> Result<kdf::Salt, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let bytes = <&[u8]>::deserialize(deserializer)?;
+        kdf::Salt::from_slice(bytes).map_err(de::Error::custom)
+    }
+}


vlmutolo:
@brycx I have a long weekend, and I'm picking up this draft PR again, along with (hopefully) the CI hardening work. Do you think that it would be better to put the serde implementations next to the types for which they're implemented, or better to put all the serde stuff in one file?

brycx:
> Do you think that it would be better to put the serde implementations next to the types for which they're implemented, or better to put all the serde stuff in one file?

I see you've updated the PR description to moving it into `typedefs`. This sounds good to me. Honestly, I believe this is something you're better suited to decide anyway, as you're also implementing it.

vlmutolo:
Yeah I'm experimenting with a few different ways of implementing this. I think moving it into typedefs is working out. I was going to ping you when it settles to take a look at decisions involved. 

And yes, before this goes out of draft status, I'll merge master into this and rebase the commits to something more organized.

vlmutolo:
Two things that may warrant special attention:

1. I implemented `TryFrom` on a couple types as a thin layer over their "from bytes" implementations. There aren't any tests for the `TryFrom` implementations specifically; it felt like I was just re-testing the code that it calls in the one-line implementations.
2. I implemented most of the `serde`-specific tests in the macro code. Not sure if it's worth testing the higher-level constructions individually. They're all basically the same.

Cargo.toml
```diff
@@ -21,6 +21,7 @@ subtle = { version = "^2.2.2", default-features = false }
 zeroize = { version = "1.1.0", default-features = false }
 getrandom = { version = "0.2.0", optional = true }
 ct-codecs = { version = "1.1.1", optional = true }
+serde = { version = "1.0.124", optional = true }
```
brycx, line 24: There seems to be missing an actual `serde` feature-definition in the `[features]` section of this file.

src/typedefs.rs
```diff
@@ -85,6 +85,59 @@ macro_rules! impl_normal_debug_trait (($name:ident) => (
...
+            std::convert::TryFrom::try_from(bytes.as_slice()).map_err(serde::de::Error::custom)
+        }
+    }
+));
+
+#[cfg(test)]
+#[cfg(feature = "serde")]
+macro_rules! test_serde_impls (($name:ident, $gen_length:expr) => (
+    #[test]
+    fn test_serde_serialized_equivalence_to_bytes_fn() {
```
brycx, line 124: We have to have at least one test in the `Tests` workflow compile with the currently-missing `serde` feature, such that these tests are run in CI.

brycx (changes requested):
This looks great so far! Just two important additions missing, mentioned in comments.

> I implemented TryFrom on a couple types as a thin layer over their "from bytes" implementations. There aren't any tests for the TryFrom implementations specifically; it felt like I was just re-testing the code that it calls in the one-line implementations.

This seems fine to me. `TryFrom` should be implicitly tested through testing `Deserialize` AFAICT.

> I implemented most of the serde-specific tests in the macro code. Not sure if it's worth testing the higher-level constructions individually. They're all basically the same.

The way you tested the `PasswordHash` separately looks good to me. The rest are very much the same. I don't see anything that needs changing here.

Cargo.toml
vlmutolo: I could be misunderstanding this, but I don't think you need to explicitly write down a feature name. Optional dependencies automatically generate features with the same label. Check out [this section of TRPL](https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies). Thoughts?

src/typedefs.rs
vlmutolo: I think `serde` is the first non-default optional feature added to Orion. Maybe the right way forward is to have an optional feature called `all` or `extra` that will explicitly enable all extra features, including `serde` in this case.

The counterpoint to this is that we don't have any other non-default optional features at the moment, and this could be an example of premature generalization. I could go either way.

For now, for testing, I'll add a new test that's just like normal, except with `serde` enabled. In the future, if we want to add more optional features, maybe we replace that test with something like `cargo test --features all`.

Cargo.toml
brycx: You are completely right. I didn't know about this. In fact, given that:
> "Note: A feature in the [feature] table cannot use the same name as a dependency."


this seems to be just right.

brycx (approved):
LGTM, nice work!

Seems the lints are complaining because `x25519` imports `TryFrom`, but the path to this trait is fully qualified in `typedefs`. If you have the time, it would be nice to have all Action workflows pass before merging. 

If possible, I think it would be fine to exclude this lint in the `typedefs` section of `impl TryFrom`, since we might import `TryFrom` other places in the future, where the `typedefs` macros are also used.

vlmutolo:
If you have the time, it would be nice to have all Action workflows pass before merging.Yep! I'll get these lint's passing before I merge.

brycx:
> Ok, good point. I think it would make sense to test serde with alloc enabled, then. At the moment, the deserialization always allocates.

Good idea. Probably, we'll end up adding some `alloc` imports in the `serde` code. `serde` itself also defines an `alloc` features, so this should probably be enabled if orion is also compiled with `alloc` feature enabled.

brycx:
Re: 
> "Do you want to do this in scope of this issue, or rather a different one?"

I've created a new issue for this.

vlmutolo:
Note that I had to make a small modification in the x25519 code to get the warnings to go away. The only change was that I fully qualified an import (`TryFrom`) instead of using a `use` statement as it was before.

The reason I had to make this change is that something in the x25519 module was importing a macro that fully qualified the `TryFrom` trait, and it was throwing a warning about unnecessary full qualification.

PR #194, opened 2021-03-27
remove erroneous non-constant-time warning

vlmutolo:
The PasswordHash docs were not updated to remove the warning about its base64 encoding and decoding operations being non-constant-time. Since the move to ct-codecs, they are now constant-time and the warning are no longer needed.

This should probably merge before v016.


brycx (approved):
Great catch! Thought I had removed it from the module docs, don't know how I missed it.

PR #197, opened 2021-04-27
X25519 using fiat-crypto Curve25519 field arithmetic

brycx:
TODO:
- [X] Decide which exact tests, from Wycheproof, should be expected to fail. Should align with the RFC.
- [X] Zeroizing `Drop` impl for `FieldElement`? -> `FieldElement` does not hold any secret bytes directly from `Scalar` so it seems fine to leave as-is.
- [X] If `Neg` for `FieldElement` is not used because it is really only used through the fiat-crypto provided, then remove it.
- [X] Documentation for all functions that do not have it yet
- [X] Should `FieldElement::as_bytes()` mask the MSB as is done in `from_bytes()`? -> No, the masking is only needed when decoding an arbitrary byte-array into a `FieldElement`. `as_bytes()` encoding is handeled by fiat-crypto.
- [X] Investigate if a cleaner approach is applicable for `invert()` for `FieldElement` -> Nothing convincing has come up so far. We leave this be for now, and can potentially re-investigate later on if it becomes an issue to review.
- [X] Decide on whether to implement `MulAssign` for `FieldElement` or not (what about other `*Assign` ops?) -> Same argument as with finding a cleaner approach for `invert()`.
- [X] `Scalar` should not be exposed in the public API, but instead a mid-level type that is wrapped by a `SecretKey`-like type
- [X] Missing error-handling 
- [X] General and differential fuzzing targets
- [X] Create a new issue for the high-level interface for X25519
- [X] `SharedSecret` newtype (can't use existing macro since a `SharedSecret` should not have a `generate()` function (maybe take same approach as with `construct_public`))


vlmutolo:
Looking through the code (in a cursory way), I'm glad that you went with auto-generated implementations of the low-level math. There's actually plenty of complexity left to take on by hand even using the generated, proven code.

brycx:
> There's actually plenty of complexity left to take on by hand even using the generated, proven code.

Yeah,

brycx:
> There's actually plenty of complexity left to take on by hand even using the generated, proven code.

Indeed. The code on top of fiat's already requires plenty to review and test.

PR #202, opened 2021-05-17
[Draft] restrict permissions for GitHub actions

vlmutolo:
This commit introduces two changes. First, the actions are changed to only have
read access to repositories. Second, we specify that GitHub should persist the
authorization token for write access to a repository on disk (see the option
`persist-credentials: false`).


vlmutolo:
For reference, I'm pinning all actions to their SHA1 hash. I investigated how often releases are cut for the various actions we use, and it's no more frequent than about once per month on average. That seems acceptable, even if we start failing CI when a new release is detected for an action.

brycx:
This looks great!

I don't know if it's always been there and when I made the original issue, but I just found an option to set the default permissions for workflows in the settings of the repository. It was read+write and changed it to read-only. I think it's fine to still explicitly state that we want read-only in each workflow. Having the default just adds the extra fail-safe in case we forget to add it to new ones.

> Second, we specify that GitHub should persist the authorization token for write access to a repository on disk (see the option
persist-credentials: false)

Did you mean _should **not** persist authorization token for write access_? I think the idea should be to avoid having it persist such that another step from a workflow can read a potential write-access token, even though the step should only be read. (Ref. https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

The rest looks perfect. And if the Dependabot updates for actions can be added (I mentioned in the issue) then we don't have to worry about how often they cut releases and check are just automatic for updates.

vlmutolo:
Oops! Good catch with the commit message. I reworded it. And like I said on the issue comment, Dependabot seems like a great solution for this. I'll look into it over the weekend.

brycx:
I've taken a look through this PR. Except for maybe updating the SHA commit refs for the Actions, is there anything blocking this from  moving out of draft and being ready to merge @vlmutolo? I'd be happy to update the remaining parts and get this merged.

PR #206, opened 2021-06-10
Add `is_empty()` to newtypes

### LINKED ISSUE #205: Add `is_empty()` to public API of all newtypes that have implemented `len()`
Clippy now warns on

```
struct `Password` has a public `len` method, but no `is_empty` method
```

IIRC all newtypes either don't accept empty input or don't hold an empty inner value after creation (e.g the HMAC `SecretKey` accepts empty input, but pads the data so the actual `SecretKey` will always have some bytes in it). The question is, if all newtypes are `is_empty() == false`, should we even add the `is_empty()` method?

Following clippy guidelines in general seems like a good idea. If we're going to add it, I think it should be added to all newtypes,, regardless of what behavior we expect from them. Maybe we also end up adding newtypes in the future, that actually can be empty, and in that case it makes sense to have it. Apart from additional testing, I see no big drawbacks and therefor think we might as well add it.

Thoughts @vlmutolo?

brycx:
closes #205

### DIFF
--- a/src/high_level/pwhash.rs
+++ b/src/high_level/pwhash.rs
@@ -336,6 +336,17 @@ impl PasswordHash {
     pub fn len(&self) -> usize {
         self.password_hash.len()
     }
+
+    #[inline]
+    /// Return `true` if the password hash is empty, `false` otherwise.
+    ///
+    /// __NOTE__: This method should always return `false`, since there shouldn't be a way
+    /// to create an empty password hash.
+    pub fn is_empty(&self) -> bool {
+        debug_assert_eq!(self.encoded_password_hash.is_empty(), self.password_hash.is_empty(),
+                   "Both the encoded password hash and the raw hash must be non-empty or empty at the same time.");
+        self.password_hash.is_empty()
+    }
 }
 
 impl core::fmt::Debug for PasswordHash {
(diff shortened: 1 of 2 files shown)


src/high_level/pwhash.rs
```diff
@@ -336,6 +336,17 @@ impl PasswordHash {
...
         self.password_hash.len()
     }
+
+    #[inline]
+    /// Return `true` if the password hash is empty, `false` otherwise.
+    ///
+    /// __NOTE__: This method should always return `false`, since there shouldn't be a way
+    /// to create an empty password hash.
+    pub fn is_empty(&self) -> bool {
+        debug_assert_eq!(self.encoded_password_hash.is_empty(), self.password_hash.is_empty(),
```
vlmutolo, line 346: The debug assert is a good compromise here.

src/typedefs.rs
```diff
@@ -193,6 +193,18 @@ macro_rules! func_len (() => (
     }
 ));
 
+/// Macro to implement a `is_empty()` function which will return `true` if `self.len() == 0`.
```
vlmutolo, line 196: Small note: maybe "an" here instead of "a" before `is_empty`.

vlmutolo (approved):
Looks good! I like the doc line. It's short and clear. I left one small change to consider, but otherwise it looks good.

src/typedefs.rs
brycx: Nice spot! Fixed this.

Issue #208, opened 2021-09-24
BLAKE2b return type with more protections


brycx:
While implementing a XChaCha20-BLAKE2b construct, as part of the PASETO v4 spec, I had to use BLAKE2b to [derive encryption and authentication keys](https://github.com/brycx/pasetors/blob/83a76eb96f6e4353cfc8270c7ca2646d2888d62e/src/version4.rs#L162). That meant dealing with the public/unprotected `Digest` newtype, holding sensitive bytes, but without the newtype having a zeroizing `Drop` implementation. Already, the `Digest` type is somewhat awkward in terms of BLAKE2b, since it's not like the `Tag` used by HMAC, that offers more protections and also results in the following documentation:

```
//! - When using Blake2b with a secret key, then the output can be used as a
//!   MAC. If this is the intention, __**avoid using**__ [`as_ref()`]
//!   to compare such MACs and use instead [`verify()`], which will compare
//!   the MAC in constant time.
``` 

Implementing the XChaCha20-BLAKE2b construct has lead me to think that the current protections offered by the return type of BLAKE2b are insufficient. So far, the following approach seems the most reasonable to me:

Add a `Drop` to the `Tag` newtypes and replace `Digest` with `Tag` in BLAKE2b. This, however, requires us to remove the auto-derive of the `Copy` trait in the `construct_tag` macro, meaning that all tag newtypes no longer implement `Copy`, only `Clone`. `Tag` seems to be the best-fitting newtype for BLAKE2b, as it's used with other MACs in Orion.

I don't see a big problem with removing `Copy` from tag, as long as `Clone` is still available. `Drop` could also be implemented for the `Digest` currently used in BLAKE2b, resulting in "less breaking" changes, but that also implies removing `Copy` for all other public newtypes.

In any case, it seems like a good idea to me, to change the return type of BLAKE2b. Sure, it can be used as a hash function only, but if it can also be used as MAC, should the return type then not represent the case where most protection is needed, to increase misuse-resistance? 

cc @vlmutolo

vlmutolo:
The rationale makes sense to me. The only concern I'd have is that returning a Tag from blake might confused users expecting a Digest. I imagine that the majority of people using blake aren't using the keyed version. That said, though, I also imagine that most people looking for a hash would use the Orion::hash module. 

Could we have hazardous::blake return a Tag, but then have the hash module return a Digest?

brycx:
[...]

PR #209, opened 2021-09-28
Separate interfaces for BLAKE2b in keyed and non-keyed mode

brycx:
see #208 

TODO:
 - [ ] Update fuzzing targets

NOTE: The duplication of testing code from `StreamingContextConsistencyTester` in `mac::blake2b` should not be redundant, since it tests using different secret key and sizes.


brycx:
@vlmutolo The only question that remains from my side on this change is whether to re-export `Tag` from `hazardous` as we do in `orion::hash` with `Digest`? We could avoid having another newtype conversion, but the current high-level sets a strict limit for the size that can be expected from the `Digest` (instead of the `1..=64` range in `hazardous`). What do you think?

vlmutolo (approved):
These changes seem like they're good to go. Only small comment would be one of documentation. Maybe there should be a line added to the top of the high-level hash documentation that says something like:

> If you're looking for a keyed hash, see the `mac` module.

vlmutolo:
@brycx I don't think it would be a big issue to re-export `Tag`. The high-level module's `authenticate` and `authenticate_verify` functions still return a `Tag` with a set length that the user can't modify. So long as the user sticks with those functions from the module — and I don't see an obvious way around using them — then the user will be fine, no?

I'm trying to think of a situation where someone would want to generate their own `Tag` from a slice and then somehow use it to authenticate something, but I can't come up with anything realistic. I think it's fine to re-export.

brycx:
> So long as the user sticks with those functions from the module — and I don't see an obvious way around using them — then the user will be fine, no?

I agree. It was simply the tighter restrictions on the newtype that maybe could've made sense. But I couldn't think of a convincing argument.

> I'm trying to think of a situation where someone would want to generate their own Tag from a slice and then somehow use it to authenticate something, but I can't come up with anything realistic. I think it's fine to re-export.

Indeed. The underlying `secure_cmp()` would fail either way, if the `expected` had a longer or shorter tag than 32 bytes. I've added another error condition to the docs, and a check in the `authenticate_verify` function for clarity on this expectation.

PR #225, opened 2021-10-16
Convert uses of assert!(a == b) to assert_eq!(a, b) where possible

### LINKED ISSUE #210: Convert uses of `assert!(a == b)` to `assert_eq!(a, b)` where possible
There seem to be several places in the codebase, where it's possible to use `assert_eq!(a, b)` instead of the, IMO, less readable `assert!(a == b)`. This conversion is also suggested by my IDE (IntelliJ).

touilleMan:
hi ;-)

fix #210

I've used the following regex to do the change:
```shell
sed -i -E 's#assert!\((.*) == (.*)\);#assert_eq!\(\1, \2\);#' src/**.rs
```

Then I had to manually revert some incorrect changes: the `debug_assert!(swap == 1 || swap == 0);` and assert on `u64x4::U64x4`/`u32x4::U32x4` types which doesn't implement `Debug` trait

### DIFF
--- a/src/hazardous/stream/chacha20.rs
+++ b/src/hazardous/stream/chacha20.rs
@@ -433,7 +433,7 @@ mod public {
             let keystream2 =
                 hchacha20(&SecretKey::from_slice(&[1u8; 32]).unwrap(), &[0u8; 16]).unwrap();
 
-            assert!(keystream1 != keystream2);
+            assert_ne!(keystream1, keystream2);
         }
 
         #[test]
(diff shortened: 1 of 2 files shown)


brycx:
Hi @touilleMan 

Thanks for doing this, much appreciated! LGTM so far, but we'll have to wait for the CI to clear it as well. Did you try to apply the regex to files in the `/test` dir as well? IIRC, we have some candidates there as well, if it's not too much trouble.

touilleMan:
Thanks for the fast reply !

I've updated the PR to also modify the tests folder 😃

Issue #227, opened 2021-10-17
integrate with standard `Read`/`Write` IO traits


vlmutolo:
## Summary
The Rust ecosystem heavily integrates with `std`'s `Read` and `Write` traits. It would be helpful to interoperate with these traits, especially in the context of the `aead` and `hash` modules.

### Hashing
```rust
use orion::hash::{self, Digest};
use std::io::Cursor;

let data = Cursor::new(vec![0, 1, 2, 3, 4]); 
let digest: Digest = hash::from_reader(&mut data);
```
### AEAD
```rust
use orion::aead;
use std::io::Cursor;

let data = Cursor::new(vec![0, 1, 2, 3, 4]);
let secret_key = aead::SecretKey::default();

// Encrypt from a `Read` type.
let mut encrypted_data = Vec::new();
aead::seal_copy(&secret_key, &mut data, &mut encrypted_data)?;

// Decrypt into a `Write` type.
let mut decrypted_data = Vec::new();
aead::open_copy(&secret_key, &mut encrypted_data, &mut decrypted_data)?;
```

I'm definitely open to suggestions on the API described here. Particularly, I'm not sure if we'd want convenience wrappers around the `*_copy` functions for the AEAD case. The copy functions are fairly general, and will allow for people to allocate buffers how they want. But there's probably a case to be made to provide a simpler type that just outputs a `Vec` of encrypted/decrypted data.

---

In any case, the real advantage of implementing these APIs is that someone could do, for example, the following:

```rust
use std::fs::File;
use orion::hash;

let file = File::open("filename.txt").unwrap(); // may want to use a BufReader (?)
let digest = hash::from_reader(file).unwrap();
```

And it would work as expected. The big deal here is that large files should be read and hashed in pieces, which currently requires reaching into the `hazardous` module and using the "streaming interface" (the `update` method). This looks something like the following.

```rust
use orion::hazardous::hash::blake2b::{Blake2b, Hasher, SecretKey};

// Figuring out the "right" size isn't something users should have to do.
let mut state = Blake2b::new(None, 32)?;
let mut reader = File::open("filename.txt")?;
let mut buf = Vec::new();

while let 0 < reader.read(&mut buf) {
    state.update(buf.as_slice())?;
    buf.clear();
}
let digest = state.finalize()?;
```

So it's longer, has a little more room for user error, and in general we probably just want to support the existing IO traits.

I already started the implementation, and I'll post a draft PR soon. 

[...]

PR #228, opened 2021-10-17
impl standard IO traits for hashing

vlmutolo:
See #227.

## TODOs
- [ ] Hashing
    - [x] implement from_reader for BLAKE2b
    - [ ] implement from_reader for sha2 (if we feel like we need to)
        - [ ] sha256
        - [ ] sha384
        - [ ] sha512
- [ ] AEAD
    - [ ] implement seal_copy for XChaCha20Poly1305
    - [ ] implement open_copy for XChaCha20Poly1305
    - [ ] Other, non-high-level-interface methods (?)
- [x] General
    - [x] decide whether to keep or change all these function names, like from_reader, seal_copy, open_copy

---

EDIT: The scope of this PR has changed. We were going to try to fit the Read/Write impls for all of Orion's types into this PR, but then decided to split it up. #227 will keep track of all the impls.

Now this PR implements `std::io::Write` for Blake2b and adds a convenience function `orion::hash::digest_from_reader`.

### DIFF
--- a/src/hazardous/hash/blake2/blake2b.rs
+++ b/src/hazardous/hash/blake2/blake2b.rs
@@ -65,6 +65,9 @@ use crate::errors::UnknownCryptoError;
 use crate::hazardous::hash::blake2::blake2b_core;
 use crate::hazardous::hash::blake2::blake2b_core::BLAKE2B_OUTSIZE;
 
+#[cfg(feature = "safe_api")]
+use std::io;
+
 construct_public! {
     /// A type to represent the `Digest` that BLAKE2b returns.
     ///
@@ -145,6 +148,38 @@ impl Hasher {
             Hasher::Blake2b384 => Blake2b::new(48),
             Hasher::Blake2b512 => Blake2b::new(64),
         }
+    }
+}
+
+#[cfg_attr(docsrs, doc(cfg(feature = "safe_api")))]
+/// Example: custom digest size.
+/// ```rust
+/// use orion::{
+///     hazardous::hash::blake2::blake2b::{Blake2b, Digest},
+///     errors::UnknownCryptoError,
+/// };
+/// use std::io::{self, Read, Write};
+///
+/// // `reader` could also be a `File::open(...)?`.
+/// let mut reader = io::Cursor::new(b"some data");
+/// let mut hasher = Blake2b::new(64)?; // 512-bit hash
+/// std::io::copy(&mut reader, &mut hasher)?;
+///
+/// let digest: Digest = hasher.finalize()?;
+///
+/// # Ok::<(), Box<dyn std::error::Error>>(())
+/// ```
+#[cfg(feature = "safe_api")]
+impl io::Write for Blake2b {
+    fn write(&mut self, bytes: &[u8]) -> io::Result<usize> {
+        self.update(bytes)
+            .map_err(|e| io::Error::new(io::ErrorKind::Other, e))?;
+        Ok(bytes.len())
+    }
+
+    fn flush(&mut self) -> Result<(), std::io::Error> {
+        // This type doesn't buffer writes, so flushing is a no-op.
+        Ok(())
     }
 }
 
@@ -342,6 +377,26 @@ mod public {
             assert!(Blake2b::new(65).is_err());
             assert!(Blake2b::new(1).is_ok());
             assert!(Blake2b::new(64).is_ok());
+        }
+    }
+
+    #[cfg(feature = "safe_api")]
+    mod test_io_impls {
+        use crate::hazardous::hash::blake2::blake2b::Blake2b;
+        use std::io::Write;
+
+        #[quickcheck]
+        fn prop_hasher_write_same_as_update(data: Vec<u8>) -> bool {
+            let mut hasher_a = Blake2b::new(64).unwrap();
+            let mut hasher_b = hasher_a.clone();
+
+            hasher_a.update(&data).unwrap();
+            hasher_b.write_all(&data).unwrap();
+
+            let hash_a = hasher_a.finalize().unwrap();
+            let hash_b = hasher_b.finalize().unwrap();
+
+            hash_a == hash_b
         }
     }
 }
--- a/src/hazardous/hash/blake2/blake2b.rs
+++ b/src/hazardous/hash/blake2/blake2b.rs
@@ -177,8 +177,8 @@ impl io::Write for Blake2b {
         Ok(bytes.len())
     }
 
+    /// This type doesn't buffer writes, so flushing is a no-op.
     fn flush(&mut self) -> Result<(), std::io::Error> {
-        // This type doesn't buffer writes, so flushing is a no-op.
         Ok(())
     }
 }
--- a/src/hazardous/hash/blake2/blake2b.rs
+++ b/src/hazardous/hash/blake2/blake2b.rs
@@ -171,6 +171,16 @@ impl Hasher {
 /// ```
 #[cfg(feature = "safe_api")]
 impl io::Write for Blake2b {
+    /// Update the hasher's internal state with *all* of the bytes given.
+    /// If this function returns the `Ok` variant, it's guaranteed that it
+    /// will contain the length of the buffer passed to [`Write`](std::io::Write).
+    /// Note that this function is just a small wrapper over
+    /// [`Blake2b::update`](crate::hazardous::hash::blake2::blake2b::Blake2b::update).
+    ///
+    /// ## Errors:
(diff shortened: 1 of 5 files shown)


vlmutolo:
## TODOs

### Hashing
- [ ] implement `from_reader` for Argon2i
- [ ] implement `from_reader` for sha2 (if we feel like we need to)
    - [ ] sha256
    - [ ] sha384
    - [ ] sha512

### AEAD
- [ ] implement `seal_copy` for XChaCha20Poly1305
- [ ] implement `open_copy` for XChaCha20Poly1305
- [ ] Other, non-high-level-interface methods (?)

### General
- [ ] decide whether to keep or change all these function names, like `from_reader`, `seal_copy`, `open_copy`

brycx:
@vlmutolo I think it'd be best to keep it behind `safe_api` for now. We can look into if it makes sense to provide `Read`/`Write` for `hazardous`-only things later.

ATM, I think `safe_api` makes most sense. Also, with the `serde` feature, I'd like to get the new wiki section written first, to see how much the new `serde` feature complicates the feature-history of Orion.

brycx:
Is there something that blocks progression on this @vlmutolo? Should we discuss it more, maybe play around with more examples?

vlmutolo:
So I don't think anything is "blocking" this exactly, but there are some open design questions I have that would be good to discuss. I'll write a bit about that on the relevant issue.

brycx:
I answered in the wrong place regarding some things. Apologies. Re-posting here:

>  decide whether to keep or change all these function names, like from_reader, seal_copy, open_copy

`form_reader` seems good to me. Perhaps the naming of the functions for non-streaming AEAD could be made a bit less generic. How about `sealing_read()`/`opening_write()` or `seal_read()`/`seal_write()`?

> Do you think there should be a new io feature for this? It currently doesn't compile without default features. 

I don't think another io feature is necessary. If we just gate it behind safe-api that should be fine, just like we do with the `std:error:Error` trait.

brycx:
Edited your issue to having completed `from_reader` for BLAKE2b instead of Argon2i.

brycx:
For when the implementation is more stable, I think it would be beneficial to try and integrate the `*_from_reader()` function into the streaming API consistency testing framework, to ensure equivalence in behavior. Possibly also `Write` impl if applicable. AEAD testing framework is an obvious candidate. But these could also be a different PR.

vlmutolo:
Maybe it would be a good idea to split this PR into two: one for hashing, one for AEAD.

brycx:
> Maybe it would be a good idea to split this PR into two: one for hashing, one for AEAD.

Good idea. I'd probably do so, especially to work on the tests and not grow the PRs too large with that alone.

vlmutolo:
I tried to use something like `#[cfg_attr(docsrs, doc(cfg(feature = "safe_api")))]` on the Blake2b Write impl, but couldn't figure get the little docs warning to show up. I also looked through how Orion already uses it and found that it was added for the `serde` impls [here](https://github.com/orion-rs/orion/blob/ece9e42fb5b42fcc871614027ddb76bc9cb43fed/src/typedefs.rs#L92). However, when I render the docs with `cargo doc --features serde`, I still don't see the little bubble pop up locally.  Am I just rendering the docs wrong?

brycx:
Did you build the docs with the parameters listed in the README? The command you just posted in the comment won't enable the rendering IIRC.

vlmutolo:
Yep, that worked. Everything shows up as it should, including the feature bubble for `Blake2b`'s `Write` impl.

src/hazardous/hash/blake2/blake2b.rs
```diff
@@ -148,6 +151,38 @@ impl Hasher {
...
+///
+/// // `reader` could also be a `File::open(...)?`.
+/// let mut reader = io::Cursor::new(b"some data");
+/// let mut hasher = Blake2b::new(64)?; // 512-bit hash
+/// std::io::copy(&mut reader, &mut hasher)?;
+///
+/// let digest: Digest = hasher.finalize()?;
+///
+/// # Ok::<(), Box<dyn std::error::Error>>(())
+/// ```
```
brycx, line 171: This is a great and succinct example! One thing I'm missing is perhaps a mention of the error-handling - that `io::ErrorKind::Other` is the only error ever thrown, if updating the hashing state errors.

In addition, it's a good comment that was put within the `flush()` fn, but seems also like a great documentation candidate. Otherwise, users might wonder why we even have flush or attempt to use it. Not a big deal though.

Lastly, maybe it's worth mentioning shortly that `Write` simply wraps `update()`, so that a reader may tell that the init->update-finalize-reset logic applies here as well, as described in the module docs.

src/high_level/hash.rs
```diff
@@ -67,6 +79,27 @@ pub fn digest(data: &[u8]) -> Result<Digest, UnknownCryptoError> {
...
 
+/// Hash data from `impl Read` using BLAKE2b-256.
+///
+/// See the [module-level docs](crate::hash) for an example of how to use this function.
+/// Internally calls `std::io::copy()` to move data from the reader into the Blake2b writer.
+/// The `std::io::copy` function buffers reads, so passing in a `BufReader` may be unnecessary.
+///
+/// For lower-level control over reads, writes, buffer sizes, *etc.*, consider using the
+/// [`Blake2b`](crate::hazardous::hash::blake2::blake2b::Blake2b) type and its
+/// [`Write`](std::io::Write) implementation directly. See the Blake2b type's source code
```
brycx, line 90: Again, really great docs. I really like the convenience of linking to `std`-related functionality on the docs, like you've enabled for the `Write`. I was hoping we could also do this for the mentions of:

- `Read`
- `std::io::copy()` (since this already a full path it doesn't even need a link, just [])
- `BufReader`

Lastly, maybe "See the Blake2b type's source code [..]" could be "See the Blake2b Write implementation source code [..]" for a small bit of added clarity on where to go look. Or point to the test if that is what is meant.

src/high_level/hash.rs
```diff
@@ -67,6 +79,27 @@ pub fn digest(data: &[u8]) -> Result<Digest, UnknownCryptoError> {
...
+/// for an example.
+///
+/// Note that if an error is returned, data may still have been consumed
```
brycx, line 93: Really great catch.

src/high_level/hash.rs
```diff
@@ -67,6 +79,27 @@ pub fn digest(data: &[u8]) -> Result<Digest, UnknownCryptoError> {
...
+/// from the given reader.
+#[cfg(feature = "safe_api")]
+#[must_use = "SECURITY WARNING: Ignoring a Result can have real security implications."]
+pub fn digest_from_reader(mut reader: impl std::io::Read) -> Result<Digest, UnknownCryptoError> {
```
brycx, line 97: I'm not well-versed in `Read`-based API's, but is it normal to take ownership of the reader as is done here? The reason I'm asking is that, it is only passed as a reference to `std::io::copy()` and I don't immediately see a need to move the reader.

brycx (changes requested):
Great start on getting `io` integration for the library! I have two general things in addition to code comments:

[1] Please rebase upon `master` so that we're sure the doctests run with `--no-default-features` (#263). I believe this should be the case already though, since the function itself is gated behind `safe-api`.

[2] Since the title of the PR is "IO traits for hashing" do you plan to add `Write` impls for SHA2 primitives in other PRs?

src/high_level/hash.rs
vlmutolo: So I think that users can pass *any* type that implements Read, and that includes a `&mut R` where `R` is a reader. That's how I've done it in other projects. In that case the unique reference will be passed by value and "consumed", but the original object will be left intact.

src/high_level/hash.rs
vlmutolo: All great points. I'll make the changes.

vlmutolo:
> Please rebase upon master

👍

> do you plan to add Write impls for SHA2 primitives

That's a good question. I can definitely do the same write impl for SHA as I did for Blake. Wasn't thinking about that, but I should probably add it to this PR.

src/high_level/hash.rs
brycx: > So I think that users can pass any type that implements Read, and that includes a &mut R where R is a reader. That's how I've done it in other projects. In that case the unique reference will be passed by value and "consumed", but the original object will be left intact.

I didn't know you could do this, but just tested it and seems to work exactly as intended. Awesome!

brycx (approved):
Sorry about the delay - LGTM, great work!

PR #240, opened 2021-10-31
Use `doc_cfg` to render feature badges for feature-gated functionality

### LINKED ISSUE #238: Render feature badges for API on docs.rs
This idea was originally proposed to me in https://github.com/brycx/pasetors/issues/29 (credit to @not-my-profile).

This will add nice badges to APIs that require specific features, which might make it easier to navigate Orion's APIs. This should however be tested beforehand, with screenshots, to see if the end result actually makes sense.

brycx:
fixes #238 

`RUSTDOCFLAGS='--cfg docsrs' cargo +nightly doc --no-deps --all-features --open`

Screenshots:
ss1
ss2
ss3
ss4

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,3 +48,6 @@ harness = false
 [profile.dev]
 opt-level = 1
 
+[package.metadata.docs.rs]
+all-features = true
+rustdoc-args = ["--cfg", "docsrs"]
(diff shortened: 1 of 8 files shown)


vlmutolo (approved):
I think it's great 👍🏼. I was initially worried about seeing a bunch of `safe_api` badges everywhere, but you restricted it to module boundaries, so I think that works well. And anything that makes it clearer what features you have to use is a big win.

PR #252, opened 2021-12-05
switch to crates.io version of fiat-crypto

vlmutolo:
Previously, we vendored the generated code from the fiat-crypto
project. It was brought to our attention that the officially
published fiat-crypto on crates.io is sufficiently up-to-date
to use in Orion and that there's no need to manually vendor
the code.

Additionally, using the crates.io version integrates well
with Dependabot so we can get alerts when our version of fiat-crypto
is out of date. This was a challenge we would have had to solve
with the vendored code that we now get solved for free.


vlmutolo:
I left the module names the same and did the import with `use fiat_crypto::curve25519_64 as fiat_curve25519_u64`. @brycx let me know if you think it would be cleaner to just update all the module names. I personally don't have a strong opinion either way.

brycx:
Rebasing upon master should fix the lint/deny actions previously failing.

brycx:
The new release of the fiat-crypto crate should now provide no_std.

brycx (approved):
LGTM.

PR #261, opened 2021-12-22
x25519 (internal): Remove mod-level allows

brycx:
These are no longer needed, since we moved to the `fiat-crypto` crate and can be deleted.


vlmutolo (approved):
LGTM

Issue #257, opened 2021-12-13
CI: Switching from `cargo-audit` to `actions-rs/audit-check`


brycx:
The README of `cargo-audit` mentions you should use `audit-check` directly, but it seems to me the repository isn't being updated currently. Last commit was March 2020 and several fixes pend review/merging for new releases. Let's keep an eye on this audit action for now.

A nice pro for switching to `audit-check` is it will open issues for any RustSec advisories.

https://github.com/RustSec/rustsec/tree/main/cargo-audit
https://github.com/actions-rs/audit-check

brycx:
This is the underlying reason as to why issues are not opened as mentioned in #251

vlmutolo:
> A nice pro for switching to audit-check is it will open issues for any RustSec advisories.

That's really cool.

brycx:
> I assume it notifies based on what dependencies you have in your tree and their versions (?).

Precisely. Just the direct dependencies in our Cargo.toml file.

PR #262, opened 2021-12-24
CI: Add `audit-check` to Actions workflow

### LINKED ISSUE #251: GH Security alerts may not work as expected
The Security Audit workflow fails because af dev-dependency of criterion is no longer maintained, and a corresponding issue has been filed with RustSec.

I would've assumed GH would've reported this as well, so we should figure out why not and how we'd enable this.

brycx:
closes #251, #257

This adds the `audit-check` action to our workflow. This should create issues if vulnerable dependencies show up during the run, instead of just the current action failing and sending me an email. It should also fix #251.

We will keep the previous cargo audit check as well because, as mentioned in #257, it seems unclear how much support this `audit-check` action is receiving.

### DIFF
--- a/.github/workflows/audit_check.yml
+++ b/.github/workflows/audit_check.yml
@@ -1,0 +1,17 @@
+name: Security audit (audit-check)
+permissions:
+  contents: read
+
+on:
+  schedule:
+    - cron: '0 0 * * *'
+jobs:
+  audit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579
+        with:
+          persist-credentials: false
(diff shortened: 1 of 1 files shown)


vlmutolo (commented):
Looks good. Just one question: this workflow requires the GitHub token (I'm assuming) it has to be able to open issues with the repository (?).

Are there restrictions we should be putting on this token?

Maybe we should restrict it to only write access for the issues, as shown [here](https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/).

brycx:
> Maybe we should restrict it to only write access for the issues, as shown here.

Great spot! We should definitely include write-access for the token, otherwise it can't open the issues.

vlmutolo (approved):
Looks good! So are the token permissions deny-by-default?

brycx:
According to the GitHub article, they should be AFAIU:

> "Any permission that is absent from the list will be set to none."

PR #263, opened 2021-12-24
Feature-gate doc examples to enable doctests on `--no-default-features` builds

### LINKED ISSUE #254: feature-gate tests that use `generate`
**Description**
Tests currently fail with `cargo test --no-default-features`.

**Additional information**
It seems like this is happening because lots of the tests generate a random key. We could probably just feature-gate all those tests behind `getrandom` or maybe `safe_api`. Thoughts?

- Orion features selected: `default-features = false`
- Orion version: `723d9ef11c2b4e5f2f7746ffa9392c7cea5d8dbf`
- Rust version: `$ rustc -V : rustc 1.56.0 (09c42c458 2021-10-18)`

We should probably also make sure the CI is testing no "no enabled features" case.

brycx:
closes #254 

Add feature-gating to doctests that make use of `safe-api` functionality. Additionally, remove `--tests` flag from `--no-default-features` builds in CI, in order to test that doctests run successfully on `no-std` builds.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -57,7 +57,7 @@ jobs:
         uses: actions-rs/cargo@844f36862e911db73fe0815f00a4a2602c279505
         with:
           command: test
-          args: --no-default-features --tests
+          args: --no-default-features
       
       - name: Test debug-mode, alloc feature
         uses: actions-rs/cargo@844f36862e911db73fe0815f00a4a2602c279505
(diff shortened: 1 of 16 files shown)


vlmutolo (approved):
So this seems simple enough. Looks good to me.

I assume that the following line being left out of the `cfg` is just to get the doctests to compile and pass (?).

```rust
Ok::<(), orion::errors::UnknownCryptoError>(())
```

brycx:
> I assume that the following line being left out of the cfg is just to get the doctests to compile and pass (?).

Exactly. When doctests like these are pre-processed and the `fn _doc_test_whatever` function is inserted before the actual doctest, we need the return type as part of the function, regardless of whether or not `safe-api` is enabled. Otherwise, it'd end up being:

```rust
fn _doctest_() -> R<T, E> {
#[cfg(feature = "safe-api") {
 [..]
 
 # Define E
}
 # No definition of E
}
```

PR #265, opened 2021-12-27
Specify MSRV with `rust-version` field in Cargo.toml

### LINKED ISSUE #250: Adding "rust-version" to Cargo.toml specifying MSRV
Should be possible to specify MSRV in the Cargo.toml file, in addition to the current approach of badge and explicit mention in README. Bumps MSRV to `1.56.0`

[link](https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field)

brycx:
closes #250


vlmutolo (approved):
Yep, looks good. I'm definitely a fan of having an explicit, in-code MSRV instead of having it just stated in the README.

PR #269, opened 2022-01-30
Capitalize crate name

brycx:
Also additional spelling corrections. I took care of the wiki a couple of days ago.


vlmutolo (approved):
This all looks good. I assume you did something like `rg orion` to find the places that needed to be fixed? (Can we assume these changes are exhaustive?)

brycx:
> This all looks good. I assume you did something like rg orion to find the places that needed to be fixed? (Can we assume these changes are exhaustive?)

Except for the license notice, yes, they should be. I just used the search that Intellij provided, covering all files.

PR #270, opened 2022-01-30
bench: Fix BLAKE2b benchmark

brycx:


vlmutolo (approved):
Depending on how pedantic you want to be, you might consider making the commit message say something like "format" instead of "fix". (As far as I can tell, this seems like a `cargo fmt`). That kind of thing helps people ignore this patch down the road.

brycx:
> Depending on how pedantic you want to be, you might consider making the commit message say something like "format" instead of "fix". (As far as I can tell, this seems like a cargo fmt). That kind of thing helps people ignore this patch down the road.

I named the first fix because it couldn't compile, because the module path to BLAKE2 was wrong. But yeah, perhaps instead of "fmt" I should write "clippy fmt" in the future.

vlmutolo:
It's funny: I stared at the diff for at least a minute trying to figure out if something was different. I missed the `blake2` in the very beginning. So, yep, all good, please ignore my previous comment.

Issue #293, opened 2022-08-08
MSRV `1.52.0` fails because of bad version of `plotters`


brycx:
cc @vlmutolo 

See #289

PR #294, opened 2022-08-10
derive debug and partialeq trait for blake2b hasher enum

black-eagle17:
This PR aims to solve this issue: https://github.com/orion-rs/orion/issues/278.

### DIFF
--- a/src/hazardous/hash/blake2/blake2b.rs
+++ b/src/hazardous/hash/blake2/blake2b.rs
@@ -115,14 +115,7 @@ impl Blake2b {
     }
 }
 
-<<<<<<< Updated upstream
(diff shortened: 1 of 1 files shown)


brycx (changes requested):
Thanks a lot for picking up this issue!

There seem to be some git merge lines mixed in, specifically those starting with </>/=. If you remove these, then it should be ready to merge (also removing the duplicate derive - final change should be one line addition total).

brycx:
@black-eagle17 I want to do a patch release due to #293, so I've gone ahead and fixed this PR up. Hope that's alright.

black-eagle17:
Thank you. I didn't have the time yet.

PR #309, opened 2022-12-02
(X)ChaCha20/(X)ChaCha20-Poly1305: Do not copy plaintext to `dst_out` before all needed keystream blocks are guaranteed to be produced

### LINKED ISSUE #308: Misuse issue in (X)ChaCha20 and (X)ChaCha20-Poly1305 APIs
During a review of parts of the codebase, I've come across a rather unfortunate misuse possibility in ChaCha20, which also affects ChaCha20-Poly1305, but much more limited therein.

#### (X)ChaCha20

In `orion::hazardous::stream::chacha20::encrypt()` we allow the user to specify the `initial_counter`, that the ChaCha20 state should be initialized with (also known as the block counter). If this `initial_counter` is too large, meaning would overflow when adding 1, then we return an error.

https://github.com/orion-rs/orion/blob/1ee15eb806dd78dc1b10dece4112b2cf68d7494e/src/hazardous/stream/chacha20.rs#L303

The problem is, that before this error is returned, we have already copied the plaintext to the output destination `dst_out` (in order to later XOR it with the keystream). 

https://github.com/orion-rs/orion/blob/1ee15eb806dd78dc1b10dece4112b2cf68d7494e/src/hazardous/stream/chacha20.rs#L347

If the `initial_counter.checked_add(1)` returns `Err()` the plaintext bytes that were __supposed__ to be XORed with the next keystream bytes, remain in plaintext in the output buffer.

```rust
    let k = SecretKey::generate();
    let n = Nonce::from_slice(&[0u8; 12]).unwrap();
    let ic: u32 = u32::MAX - 1;

    let text = [b'x'; 128 + 4];
    let mut dst_out = [0u8; 128 + 4];

    let err = encrypt(&k, &n, ic, &text, &mut dst_out).unwrap_err();

    assert_eq!(&[b'x'; 4], &dst_out[dst_out.len() - 4..]);
```

While we do return an error, and a user should not use the output of `encrypt()` if it returns an error, they might reasonably expect that `dst_out` not contain a copy of some plaintext bytes.

This issue also applies to XChaCha20, as it uses `chacha20::encrypt()`.

#### (X)ChaCha20-Poly1305

The impact on (X)ChaCha20-Poly1305 is more limited, as the user cannot specify the `initial_counter`. However, the plaintext is encrypted with an initial block counter of 1, because the first is used for the Poly1305 key generation. Thus, plaintext bytes may be left in the output buffer if the length of the plaintext is one block less than the maximum (max is `2^32-1 * 64` bytes), which is very unlikely to happen in real-world use.

For the `streaming` implementation, one block less than the required amount for (X)ChaCha20-Poly1305, is required to hit this, as the initial counter there used internally starts with 2, for plaintext bytes.

#### Solution

The obvious solution here seems to be, to only copy anything to the output buffer that is ciphertext. Thus, we change the underlying XOR macro usage in ChaCha20 and can take the same approach when processing plaintext bytes in AEAD. In AEAD, we could maybe even get a bit of speedup as we could then encrypt and authenticate the ciphertext block individually, instead of making two passes over the plaintext and ciphertext bytes (first encrypt, then authenticate).

We could also discuss removing the option to set `initial_counter` for (X)ChaCha20, in a future major release, as there seems to be no real use for it.

cc @vlmutolo

brycx:
closes #308

### DIFF
--- a/src/hazardous/aead/chacha20poly1305.rs
+++ b/src/hazardous/aead/chacha20poly1305.rs
@@ -180,6 +180,7 @@ pub fn seal(
     }
 
     let ad = ad.unwrap_or(&[0u8; 0]);
+    #[allow(clippy::absurd_extreme_comparisons)]
     if u64::try_from(ad.len()).map_err(|_| UnknownCryptoError)? > A_MAX {
         return Err(UnknownCryptoError);
     }
@@ -271,6 +272,7 @@ pub fn open(
         return Err(UnknownCryptoError);
     }
     let ad = ad.unwrap_or(&[0u8; 0]);
+    #[allow(clippy::absurd_extreme_comparisons)]
     if u64::try_from(ad.len()).map_err(|_| UnknownCryptoError)? > A_MAX {
         return Err(UnknownCryptoError);
     }
--- a/src/hazardous/aead/chacha20poly1305.rs
+++ b/src/hazardous/aead/chacha20poly1305.rs
@@ -166,20 +166,67 @@ pub fn seal(
         None => return Err(UnknownCryptoError),
     };
 
-    let mut enc_ctx =
+    let mut stream =
         ChaCha20::new(secret_key.unprotected_as_bytes(), nonce.as_ref(), true).unwrap();
     let mut tmp = Zeroizing::new([0u8; CHACHA_BLOCKSIZE]);
 
-    let pt_len = plaintext.len();
-    if pt_len != 0 {
-        dst_out[..pt_len].copy_from_slice(plaintext);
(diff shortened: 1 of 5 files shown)


brycx:
The way it is solved, the same way for `chacha20::poly1305::seal()` and `chacha20::encrypt()`, specifically the latter, also implies a fix for `aead::streaming`. This is because `chacha20::encrypt()` is the only function used on plaintext in `aead::streaming`.  The crate-internal `encrypt_in_place()` and `xor_keystream()` are used only with a combination input length and `initial_counter` we know are possible to produce.

A test has been added to detect this error in `chacha20::encrypt()` in the future, but not so for `aead::*`, as this would require is to encrypt a unreasonably large amount of data for testing, which likely wouldn't fit in the test runners memory either.

Issue #317, opened 2023-02-07
Adding key-committing AEAD implementations (CTX or UtC/RtC/HtE)


brycx:
I think it would make sense to start thinking about adding key-committing AEADs to the library. In general there seem to be two recent proposals that transform non-committing AEs into committing ones.

The first, are the UtC/RtC/HtE constructions by Bellare & Hoang in https://eprint.iacr.org/2022/268. Second, we have CTX proposed by Chan & Rogaway in https://eprint.iacr.org/2022/1260.

The most notable difference is that the UtC/RtC transforms work for either nonce-based AEADs and MRAEs, respectively. CTX only works for nonce-based AEADs. Orion currently only supports nonce-based AEADs, but if we choose to go with CTX then we'd need to implement RtC/HtE later on, if we add MRAE as well and wish to add key-commitment there as well.

On the other hand, CTX is much simpler to implement than UtC/RtC/HtE. Further, CTX commits to the entire input set (key, nonce, ciphertext, ad), whereas UtC only commits to the key and needs HtE on top, to be fully committing. For this reason, I think it would be best to start with CTX.

Questions remaining:
- Should this be behind a experimental flag?
- The intention is to eventually replace the high-level AEAD implementation to be fully committing. Should the high-level interface be modified in the first release, even if gated behind a feature flag, or not?
- What sizes for the new authentication tags should be chosen? 256-bit seems to be a reasonable choice. This is important because the security of the commitment scheme is based on the collision-resistance of the hash function used for CTX.
- Do we add key-commitment to XChaCha20-Poly1305 only or ChaCha20-Poly1305 also (I think both are good, then it makes sense to start providing key-committing implementations in the hazardous layer and move it to high-level up later on)

samuel-lucas6:
Hi, I saw your blog post on r/crypto, and that led me here. Interestingly, I'm also doing a master's and am doing my dissertation on committing AEAD, although I haven't started yet. Respect for simplifying the topic; the literature is a mess with the different terminology and overcomplicated language.

I think a patch worth considering is the modified padding fix by Loup Vaillant, which I implemented [here](https://github.com/samuel-lucas6/kcChaCha20-Poly1305). It uses free output after the Poly1305 key and can be implemented without introducing a timing difference between tag/commitment verification. However, it's not fully committing and does lead to a bit of ciphertext expansion.

[...]

PR #324, opened 2023-03-04
Experimental implementation of CTX

brycx:
see #317

- Adds new crate feature `experimental`
- Adds new module `hazardous::cae`
- Adds experimental implementation of CTX construction built on (X)ChaCha20-Poly1305-BLAKE2b

TODO:
- [x] Add tests using Wycheproof that both CTX constructions still produce the same ciphertext as the underlying AE
- [x] Add documentation and example of partitioning oracle attack as use-case


src/hazardous/cae/chacha20poly1305blake2b.rs
```diff
@@ -0,0 +1,165 @@
...
+
+/// The size of the BLAKE2b authentication tag.
+pub const TAG_SIZE: usize = 32;
+
+/// The maximum size of the ciphertext.
+pub const C_MAX: u64 = P_MAX + (TAG_SIZE as u64);
+
+#[must_use = "SECURITY WARNING: Experimental feature."]
+/// CTX ChaCha20Poly1305 with BLAKE2b-256.
+pub fn seal(
```
vlmutolo, line 45: What does marking this as `#[must_use]` do? Wouldn't this make you do `let _ = seal(...);` to avoid the warning?

vlmutolo:
Looks good to me at a glance. One thing to consider would be adding an example of what problem this solves compared to non-committing AEAD. I'm not sure if that's helpful or not for theoretical consumers of the library.

If we wanted to add an example, I thought the [partitioning oracle attack](https://www.usenix.org/conference/usenixsecurity21/presentation/len) was relatively straightforward.

If we ever want to change over the high-level API to a committing AEAD, I don't think that kind of note would be helpful since users aren't expected to really understand cryptography. But in the hazardous API, maybe it's good for people scrolling through the different options to have an understanding of the kinds of attacks this can prevent.

src/hazardous/cae/chacha20poly1305blake2b.rs
brycx: This is actually just because I copied from the `ChaCha20-Poly1305` directly and wasn't sure what I wanted to do with this. I think as long as the `Result` is checked it avoids a warning. Perhaps just keep it as it was before, a warning for experimental has been put in the module anyways.

brycx:
> If we wanted to add an example, I thought the [partitioning oracle attack](https://www.usenix.org/conference/usenixsecurity21/presentation/len) was relatively straightforward.

If you mean add it in description and provide the link, then yes, completely agree! Though if you were thinking of an actual code example, I don't know if this would be too big/complex to have as an example.

> If we ever want to change over the high-level API to a committing AEAD, I don't think that kind of note would be helpful in the high-level docs since users aren't expected to really understand cryptography. But in the hazardous API, maybe it's good for people scrolling through the different options to have an understanding of the kinds of attacks this can prevent.

Indeed. We should add this, but still mention in high-level that it is fully-committing. Though, if somebody searches `hazardous` for this, they probably know what they are looking for.

brycx:
@vlmutolo was this the kind of documentation addition you were thinking of?

PR #473, opened 2025-04-28
make `alloc` and `serde` compilation more robust

joseluis:
This fixes compilation that didn't succeed before like this: 
```sh
cargo b --features serde,alloc --no-default-features
```

And by auto-enabling `alloc` with serde it now works also like this:
```sh
cargo b --features serde --no-default-features
```

And it keeps working as before, like this:
```sh
cargo b
cargo b --features serde
cargo b --no-default-features
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -51,6 +51,9 @@ jobs:
       - name: Test debug-mode, no-default + alloc feature
         run: cargo test --no-default-features --features alloc --tests
 
+      - name: Test debug-mode, no-default + serde feature (enables alloc)
+        run: cargo test --no-default-features --features serde --tests
+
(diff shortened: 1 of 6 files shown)


brycx (commented):
Thank you, @joseluis!

This looks good to me and I don't think the change should pose any issues, especially since the `serde` dependency itself uses `alloc` when default-features are disabled.

It think we should add this to the test-suite in `test.yml`, to make sure this feature-combination won't regress in the future. Could I bother you to add this, @joseluis? If not that's totally fine - I'll merge this later and add it myself.

brycx (approved):
Thank you again, @joseluis! Also for adding the extra test logic, this is very helpful! (apologies for the late 2nd review)

PR #477, opened 2025-05-05
Added unit test for edge cases on x25519 keys

RoscoeEH:
This introduces a unit test for handling edge cases related to private keys in the x25519 algorithm. The test checks several edge cases for private keys, ensuring the implementation correctly handles these unusual but valid cases.

Test Scenarios Added:
- All Zeros: A private key with all zero values, which is clamped in RFC 7748, should be accepted.
- All Ones: A private key with all ones, which is clamped, should be accepted.
- Alternating Bits: A private key with an alternating bit pattern, which is an arbitrary pattern, should be accepted.
- First Byte Set: A private key where only the first byte is set, should be accepted.
- Last Byte Set: A private key where only the last byte is set, should be accepted.

### DIFF
--- a/src/hazardous/ecc/x25519.rs
+++ b/src/hazardous/ecc/x25519.rs
@@ -891,23 +891,31 @@ mod public {
     #[test]
     fn test_privatekey_edge_cases() {
         // all zeros - clamped in RFC 7748, should be accepted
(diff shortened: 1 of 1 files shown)


RoscoeEH:
I went back and changed some things to pass the tests. Everything looks good now!

brycx (approved):
LGTM.

Thank you very much, @RoscoeEH! Greatly appreciated.

PR #478, opened 2025-05-12
Unit test verifying that the most significant bit of the last byte in an x25519 public key is correctly masked

RoscoeEH:
This introduces a unit test to verify the proper handling of the most significant bit (MSB) masking in public keys for the x25519 algorithm. The test checks that the `PublicKey::from_slice` function correctly masks the MSB of the final byte of the public key when it's constructed.

Test Scenarios Added:
- MSB Set in Input: A public key where the most significant bit (MSB) of the last byte is set. The test ensures the MSB is cleared during the construction of the public key.
- Correctness of Masking: Verifies that the final byte after masking matches the original byte with the MSB cleared.

This ensures that public keys conform to the RFC 7748 §5, which states that "implementations of X25519 MUST mask the most significant bit in the final byte".

### DIFF
--- a/src/hazardous/ecc/x25519.rs
+++ b/src/hazardous/ecc/x25519.rs
@@ -891,23 +891,31 @@ mod public {
     #[test]
     fn test_privatekey_edge_cases() {
         // all zeros - clamped in RFC 7748, should be accepted
-        assert!(PrivateKey::from_slice(&[0u8; PRIVATE_KEY_SIZE]).is_ok(), "all-zero key");
-    
+        assert!(
+            PrivateKey::from_slice(&[0u8; PRIVATE_KEY_SIZE]).is_ok(),
+            "all-zero key"
+        );
+
         // all ones - clamped, should be accepted
-        assert!(PrivateKey::from_slice(&[0xffu8; PRIVATE_KEY_SIZE]).is_ok(), "all-ones key");
-    
+        assert!(
+            PrivateKey::from_slice(&[0xffu8; PRIVATE_KEY_SIZE]).is_ok(),
+            "all-ones key"
+        );
+
         // alternating bits - arbitrary pattern, should be accepted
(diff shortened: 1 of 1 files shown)


RoscoeEH:
One of the tests failed due to an unused extern create in src/lib.rs, I removed it for the test but that wasn't something I changed. It was part of a commit that I pulled down. You may want to add it back if it is intentional.

brycx (changes requested):
Thank you for taking the time to add this, @RoscoeEH.

Unfortunately, the test case is already covered https://github.com/orion-rs/orion/blob/1dd79f2882f051458e6a863e0d2838a1bf17f70d/src/hazardous/ecc/x25519.rs#L705 and as such is a duplicate.

You may however add the `assert_eq!(pk_zero.to_bytes(), pk_one.to_bytes());` to the existing test-case, because the existing one's check the `FieldElement` and not byte-arrays. Otherwise, I'll close this.

> One of the tests failed due to an unused extern create in src/lib.rs, I removed it for the test but that wasn't something I changed. It was part of a commit that I pulled down. You may want to add it back if it is intentional.

Are you using a recent version of cargo? I can't reproduce your issue.

RoscoeEH:
I added the portion that compares raw bits to the existing test and removed the redundant test. 

> Are you using a recent version of cargo/rustc? I can't reproduce your issue.

I can't produce it on my machine either, it is only from one of the Github automated checks that runs on pull requests.

brycx:
> I added the portion that compares raw bits to the existing test and removed the redundant test.

Perfect!

> I can't produce it on my machine either, it is only from one of the Github automated checks that runs on pull requests.

I see, misunderstood. Thanks for fixing this also, both approaches seem to check out AFAICT.

brycx (approved):
Thank you, @RoscoeEH!

Issue #433, opened 2025-02-23
Add support for scrypt


brycx:
@finfet has mentioned possibly upstreaming their scrypt implementation to Orion in https://github.com/finfet/kestrel/issues/11#issuecomment-2676648459.

I believe this would be a worthwhile addition to Orion, since the password-based key derivation functions right now are either Argon2i or PBKDF2. If one were to interpret [OWASP recommendations](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html) strictly, then a user would have to choose PBKDF2, because the Argon2id variant isn't supported right now.

scrypt could serve as a great alternative to Argon2i.

finfet:
I think this will be a great addition and it will give people access to an another scrypt implementation. I plan to make some progress on this over the next week.

PR #493, opened 2025-06-08
Add scrypt implementation #433

finfet:
Adds a rust implementation of scrypt from RFC 7914. It requires the alloc feature right now, but could probably be updated to work without it.

I'm happy to hear any feedback, or make changes as required.

### DIFF
--- a/src/hazardous/kdf/mod.rs
+++ b/src/hazardous/kdf/mod.rs
@@ -28,5 +28,10 @@ pub mod pbkdf2;
 
 #[cfg(any(feature = "safe_api", feature = "alloc"))]
 #[cfg_attr(docsrs, doc(cfg(any(feature = "safe_api", feature = "alloc"))))]
+/// Scrypt Password-Based Key Derivation Function as specified in the [RFC 7914](https://datatracker.ietf.org/doc/html/rfc7914.html).
+pub mod scrypt;
+
+#[cfg(any(feature = "safe_api", feature = "alloc"))]
+#[cfg_attr(docsrs, doc(cfg(any(feature = "safe_api", feature = "alloc"))))]
 /// Argon2i password hashing function as described in the [P-H-C specification](https://github.com/P-H-C/phc-winner-argon2/blob/master/argon2-specs.pdf).
 pub mod argon2i;
--- a/src/hazardous/kdf/scrypt.rs
+++ b/src/hazardous/kdf/scrypt.rs
@@ -1,0 +1,489 @@
+// MIT License
+
+// Copyright (c) 2025 The orion Developers
+
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+// Based off of the original Go implementation
+//
+// Copyright 2012 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+//! # About:
+//! scrypt as specified in RFC 7914
+//!
+//! # Note:
+//! - This implementation is only supported on platforms with pointer sizes of
+//! at least 32 bits.
+//! - This implementation is single threaded. Values for `p > 1` will not
+//!   benefit from the speedup of being run in parallel.
+//!
+//! # Parameters:
+//! - `n`: Parameter n must be larger than 1 and a power of 2
+//!
+//! # Security
+//!
+//! - The minimum recommended length for a salt is 16 bytes.
+//! - The minimum recommended length for a hashed password is 16 bytes.
+//! - The minimum recommended `n` is `16384`
+//! - The minimum recommended `r` is `8`
+//! - The minimum recommended `p` is `1`
+
+use alloc::vec;
+
+use crate::errors::UnknownCryptoError;
+use crate::hazardous::kdf::pbkdf2::sha256 as pbkdf2;
+use crate::util;
+
+use zeroize::Zeroize;
+
+/// scrypt r * p must be less than 2^30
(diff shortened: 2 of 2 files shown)


finfet:
I rebased and pushed so the codecov is passing now.

brycx:
Thank you so much for all the work on this, @finfet!

I think the easiest approach would be to get this scrypt implementation merged in when ready and do a follow-up PR after that, to take care of surrounding (boring) things like make the API docs fully consistent and expanded, etc.

Feel free to take up any of those if you're interested, otherwise I'll deal with those.

src/hazardous/kdf/mod.rs
```diff
@@ -26,6 +26,10 @@ pub mod hkdf;
 /// PBKDF2(Password-Based Key Derivation Function 2) as specified in the [RFC 8018](https://tools.ietf.org/html/rfc8018).
 pub mod pbkdf2;
 
+#[cfg(feature = "alloc")]
```
brycx, line 29: IIUC, this should work on both `std` and `alloc`-specific features (correct me if I'm wrong). I think we can make the conditionals just like those for `argon2i`.

Also with the added `cfg_attr` which highlights these features nicely in rendered docs. 

```suggestion
#[cfg(any(feature = "safe_api", feature = "alloc"))]
#[cfg_attr(docsrs, doc(cfg(any(feature = "safe_api", feature = "alloc"))))]
```

src/hazardous/kdf/mod.rs
```diff
@@ -26,6 +26,10 @@ pub mod hkdf;
...
+/// Scrypt Password-Based Key Derivation Function as specified in the [RFC 7914](https://datatracker.ietf.org/doc/html/rfc7914.html)
```
brycx, line 30:
```suggestion
/// Scrypt Password-Based Key Derivation Function as specified in the [RFC 7914](https://datatracker.ietf.org/doc/html/rfc7914.html).
```
nit.

src/hazardous/kdf/scrypt.rs
```diff
@@ -0,0 +1,445 @@
+// MIT License
+
+// Copyright (c) 2020-2025 The orion Developers
```
brycx, line 3:
```suggestion
// Copyright (c) 2025 The orion Developers
```
Copyright year starts when code is added.

src/hazardous/kdf/scrypt.rs
```diff
@@ -0,0 +1,445 @@
...
+                    .expect("invalid scrypt parameters");
+                let exp = case.expected_dk;
+                assert_eq!(exp, got.as_slice())
+            }
+        }
+
+        fn test_invalid_param() {
+            let mut dk = vec![0u8; 32];
+            assert_eq!(derive_key(b"password", b"salt", 1, 8, p, &mut dk).is_err());
+        }
```
brycx, line 443: There's a couple of things wrong in this one. I'd recommend running tests with `cargo test --all-features`, which should hit this code-path and the reason CI is still failing.

I think it would make sense to make one test for each parameter error-case, so make it easy to verify all parameters are covered. Let me know what you think.

src/hazardous/kdf/scrypt.rs
```diff
@@ -0,0 +1,445 @@
...
+            assert!(verify(&expected_dk, password, salt, n, r, p, &mut dst_out).is_ok());
+            assert!(verify(&modified_dk, password, salt, n, r, p, &mut dst_out).is_err());
+        }
+    }
+
+    #[cfg(feature = "alloc")]
+    mod test_derive_key {
+        use super::derive_key;
+
+        struct ScryptVector<'a> {
```
brycx, line 323: Can we add a link to where these test-vectors have been sourced? If the RFC, just a link to that one, if a git-repo, then a link to that with a specific commit-hash would be good.

src/hazardous/kdf/scrypt.rs
```diff
@@ -0,0 +1,445 @@
...
+    p: u32,
+    dst_out: &mut [u8],
+) -> Result<(), UnknownCryptoError> {
+    if !(n > 1)
+        || !(n & (n - 1) == 0)
+        || !(r * p < 1 << 30)
+        || !(r <= (i32::MAX as u32) / 128 / p)
+        || !(r <= (i32::MAX as u32) / 256)
+        || !(n <= (i32::MAX as u32) / 128 / r)
+        || !(usize::BITS >= 32)
```
brycx, line 246: I'd prefer if we set these maximum values as `pub const` with more clear names. That'd make them a lot easier to refer to.

src/hazardous/kdf/scrypt.rs
```diff
@@ -0,0 +1,445 @@
...
+    let r: usize = r as usize;
+    let p: usize = p as usize;
+
+    let vlen: usize = 32 * n * r;
+    let mut x = vec![0u32; 32 * r];
+    let mut y = vec![0u32; 32 * r];
+    let mut v = vec![0u32; vlen];
+    let pass = pbkdf2::Password::from_slice(password)?;
+    let blen: usize = p * 128 * r;
+    let mut b = vec![0u8; blen];
```
brycx, line 261: Should this `Vec<u8>` be `Zeroizing::new(...)`? It's the weakest hash possible of a sensitive user-supplied password, so if the performance penalty isn't outrageous, maybe it makes sense to have it zeroed after it goes out of scope?

brycx (changes requested):

src/hazardous/kdf/scrypt.rs
```diff
@@ -0,0 +1,445 @@
...
+}
+
+#[must_use = "SECURITY WARNING: Ignoring a Result can have real security implications."]
+/// scrypt key derivation function as specified in RFC 7914
+pub fn derive_key(
+    password: &[u8],
+    salt: &[u8],
+    n: u32,
+    r: u32,
+    p: u32,
```
brycx, line 237: I think we need to restrict `p` to only allow `1`, as Orion doesn't have any internal way to spin up threads which would be expected from this. The same is true for Argon2i.

src/hazardous/kdf/mod.rs
finfet: Thanks. Added.

src/hazardous/kdf/mod.rs
finfet: Fixed.

src/hazardous/kdf/scrypt.rs
finfet: Whoops, wrote that one too quickly. I fixed the test and added tests for all of the combinations.

src/hazardous/kdf/scrypt.rs
finfet: Added the link.

src/hazardous/kdf/scrypt.rs
finfet: I'm okay adding these as constants if you'd like. However, these are internal sanity checks to make sure we don't get any unreasonable values. I kept these checks because they in the port from the go implementation, but based on the discussion in the RFC, they might not even be needed.

From the RFC 7914 errata:
>  The presented limit on N was incorrectly derived from the original scrypt publication. The correct theoretical upper limit on N is 2^(128 * r) for r < 5, and 2^512 for all other values of r. Thus, the least upper bound is 2^128, which far exceeds all possible values for N in the foreseeable future, making the limit irrelevant for current implementations.

src/hazardous/kdf/scrypt.rs
finfet: I'm potentially okay with limiting this to 1, but I would prefer to just have a warning about it in the documentation. I went ahead and added a note about this.

It still produces valid results, just not as fast as if it were computed in parallel. I think it is worth allowing people to use higher p values if they need to. The Go implementation doesn't do these computations in parallel and it is still widely used. Plus,  if we block higher p values I think we're technically not RFC compliant.

src/hazardous/kdf/scrypt.rs
finfet: Added these as constants.

brycx (approved):
Everything looks good.

Thanks so much for your work on this, @finfet!

PR #509, opened 2025-07-14
fix: improve error handling in secure_rand_bytes function

reallesee:
Replace unwrap() with map_err() to handle errors from getrandom::fill gracefully, returning UnknownCryptoError instead of panicking on failure.

### DIFF
--- a/src/util/mod.rs
+++ b/src/util/mod.rs
@@ -78,7 +78,7 @@ pub fn secure_rand_bytes(dst: &mut [u8]) -> Result<(), errors::UnknownCryptoErro
         return Err(errors::UnknownCryptoError);
     }
 
-    getrandom::fill(dst).unwrap();
+    getrandom::fill(dst).map_err(|_| errors::UnknownCryptoError)?;
 
     Ok(())
 }


brycx:
Hi @reallesee,

Thanks to for taking the time to report this. I have a few things I'd like to consider, before changing this (although it's relatively simple LoC-wise, the impact is not so much).

The `unwrap()` was introduced in 2019, and due to my lack of documentation, I'm not sure I remember correctly the reason behind the change. Taking a guess, it was because it was hard to define what an error from calling CSPRNG could constitute. Do we need to wait for entropy to be available? Did we generate _some_ bytes but not enough (and will users handle this correctly). From what I remember, the chances of `getrandom()` actually failing (apart from unsupported platforms ofc) was relatively low, so the `unwrap()` was introduced to have a hard panic, so if it actually ever failed users should be assuming something would be wrong in general, instead of trying to recover and use some maybe good CSPRNG bytes.

It's been some time since I investigated this, and perhaps things have changed that make it worth propagating the error? Maybe you have some ideas about this.

If this is change is to accepted, it would also require changes throughout most of the library's Error/Panic docomentation sections. Most newtypes use this internally for `default()/generate()`.

PR #511, opened 2025-07-18
chacha20poly1305: add XChaCha20-Poly1305 support

reallesee:
Adds an implementation of the XChaCha20-Poly1305 AEAD that uses a 24-byte nonce instead of the standard 12-byte one. This makes it possible to safely generate random nonces without risk of collisions, which is particularly important for systems where nonce uniqueness cannot be ensured by other means


brycx (changes requested):
Did you open a MR with the wrong branch? Either way I'm closing this because the change makes no sense as it is right now and we already support XChaCha20-Poly1305.

PR #547, opened 2025-10-01
Update MSRV badge to 1.83

Sparadanni-eth:
### Description

Updated the MSRV badge from 1.81 to 1.83:
- Cargo.toml sets `rust-version = "1.83"`.
- README “Minimum Supported Rust Version” states 1.83.
- Keeping the badge in sync avoids misleading users about the minimum toolchain.


brycx (approved):
Thanks!

PR #616, opened 2026-04-28
fix: correctly zeroize value field in Drop impl instead of iterator

Olexandr88:
Calling .zeroize() on the iterator instead of self.value meant secret keys were never cleared from memory on drop

### DIFF
--- a/src/typedefs.rs
+++ b/src/typedefs.rs
@@ -129,7 +129,7 @@ macro_rules! impl_drop_trait (($name:ident) => (
             #[cfg(feature = "zeroize")]
             {
               use zeroize::Zeroize;
-              self.value.iter_mut().zeroize();
+              self.value.zeroize();
             }
         }
     }


brycx:
> "Calling .zeroize() on the iterator instead of self.value meant secret keys were never cleared from memory on drop"

This is something I'm unable to reproduce. It does zero the values. The following tests-script runs fine and verifies they are in fact being zeroed out (`orion v0.17.14`):

```rust
    use orion::hazardous::aead::chacha20poly1305::SecretKey;
    let heldptr = Box::new(SecretKey::from([0x42; 32]));
    let raw = Box::into_raw(heldptr);
    unsafe {
        std::ptr::drop_in_place(raw);
        let skbytes = std::slice::from_raw_parts(raw as *const u8, std::mem::size_of::<SecretKey>());
        assert_eq!(&skbytes[..32], &[0u8; 32]);
        println!("{:?}", &skbytes);
    }
```

If there's a working PoC for them not being zeroed I'm open to fixing it, otherwise I'll close this issue (EDIT: re-open if you got a PoC).

PR #622, opened 2026-05-16
fix: prevent panic on zero r or p in scrypt::derive_key

sashaphmn:
Passing r = 0 or p = 0 caused a runtime panic (integer division by zero) instead of returning Err(UnknownCryptoError).

Fix: add explicit r == 0 and p == 0 checks before the division-based guards.
Tests added for both cases.


brycx (approved):
Thank you, @sashaphmn! And apologies for the delay in review.

PR #641, opened 2026-07-19
feature: implement blake3 hash and keyed hash modes

### LINKED ISSUE #394: Add Blake3 support
**Summary**
Brief explanation of the feature.

Orion currently supports Blake2b hashing through

```
    let digest = orion::hash::digest(data).unwrap();
```

which uses Blake2b-256. However, Blake3 is 7x faster https://github.com/wazuh/wazuh/issues/15632.

**Basic example**
If the proposal involves a new or changed API, include a basic code example. Omit this section if it's not applicable.

I propose adding Blake3 through another function to prevent breaking old code. Maybe `orion::hash::digest_blake3`?

```

**Motivation**
Why are we doing this? What use cases does it support? Will it be used anywhere?

It would greatly increase performance of the digest function and allow for variable output sizes, such as 16 bytes and up.

### PR DESCRIPTION
**Pull request by jakub-mata**
This PR introduces the BLAKE3 hash function into the `/hazardous` module, implemented according to the [official BLAKE3 specifications](https://raw.githubusercontent.com/BLAKE3-team/BLAKE3-specs/master/blake3.pdf).
Features Included:
- hash mode: standard BLAKE3 hashing.
- keyed hash mode: Supports message authentication (MAC) use cases using a 256-bit key.
- extendable output (XOF): Arbitrary output lengths as per the spec.

Note: The "derive key" mode has been omitted from this PR to keep the scope focused on standard hashing. The underlying state machine has been structured so that key derivation can be easily added in a future PR if desired.

Testing includes:
- unit tests: located at the bottom of their respective source files.
- known-answer tests: set up to match the existing blake2 tests. The test vectors were sourced from the [b3sum repository](https://github.com/michaelforney/b3sum/blob/master/test_vectors.json) and cover both standard and keyed hashing across various chunk boundaries.

This initial implementation establishes a single-threaded baseline. Because BLAKE3's tree structure is designed for infinite parallelism, a natural next step will be adding a data-parallelism feature flag to support multi-threaded hashing for large inputs.

Resolves #394

### CODE REVIEW

**Comment by brycx** [comment]
> clippy / linting warning in a file I have not touched (I'm not sure how that happened)

You were unlucky and hit a new Rust version which introduced a new lint which made the CI fail. So no fault at all on your part. If you rebase on current `master`, this should be fixed now.


> Note: The "derive key" mode has been omitted from this PR to keep the scope focused on standard hashing. The underlying state machine has been structured so that key derivation can be easily added in a future PR if desired.

+

> This initial implementation establishes a single-threaded baseline. Because BLAKE3's tree structure is designed for infinite parallelism, a natural next step will be adding a data-parallelism feature flag to support multi-threaded hashing for large inputs.

This is a perfect approach, thank you for taking this into account!

File: tests/test_data/third_party/blake3_test_vectors.json
Comment (brycx) [inline tests/test_data/third_party/blake3_test_vectors.json:1]: Looks like you linked a forked repository in the PR description for these test vectors: https://github.com/michaelforney/b3sum/blob/master/test_vectors.json

Is there a specific reason for this and for not using the official repository: https://github.com/BLAKE3-team/BLAKE3/blob/master/test_vectors/test_vectors.json?

File: tests/hash/blake3_kat.rs [hunk 1]
```diff
@@ -0,0 +1,44 @@
...
+fn build_blake3_input(len: usize) -> Vec<u8> {
+    let mut data = Vec::with_capacity(len);
+    for i in 0..len {
+        data.push((i % 251) as u8);
+    }
+    data
+}
+
+#[test]
+fn test_blake3_kat() {
```
brycx, line 18: Please add a link to wherever these test-vectors come from, with a link to the repository which inlcudes specific commit at which it was pulled (e.g. "Taken from: https://github.com/BLAKE3-team/BLAKE3/commit/ae3e8e6b3a5ae3190ca5d62820789b17886a0038"). 

This allows us to easily check if it has changed in the future.

tests/hash/blake3_kat.rs
```diff
@@ -0,0 +1,44 @@
...
+    let key = key_str.as_bytes().to_vec();
+
+    // Iterate over the "cases" array
+    if let Some(cases) = root.get("cases").unwrap().as_array() {
+        for test_case in cases {
+            let input_len = test_case.get("input_len").unwrap().as_u64().unwrap() as usize;
+
+            // Extract the expected hex strings
+            let hash_hex = test_case.get("hash").unwrap().as_str().unwrap();
+            let keyed_hash_hex = test_case.get("keyed_hash").unwrap().as_str().unwrap();
```
brycx, line 40: Can we add a small comment that Orion doesn't implement key derivation atm? Makes it easier during auditing and mutation checks.

src/hazardous/hash/blake3/mod.rs
```diff
@@ -0,0 +1,157 @@
...
+};
+use core::cmp::min;
+
+/// Blake3 can run in different modes based on usage.
+/// Currently, only Hash and KeyedHash is supported.
+pub enum Mode {
+    /// Regular hashing
+    Hash,
+    /// Hashing parametrized with a key. Useful for MACs.
+    KeyedHash {
```
brycx: This cannot simply be a raw array. We have newtypes for secret types that need to be used in order to make sure they aren't leaked or misused.

What we need here is something similar to how we handle it for the other MACs:

```rust 
construct_secret_key! {
    /// A type to represent the secret key that BLAKE2b uses for keyed mode.
    ///
    /// # Errors:
    /// An error will be returned if:
    /// - `slice` is empty.
    /// - `slice` is greater than 64 bytes.
    ///
    /// # Panics:
    /// A panic will occur if:
    /// - Failure to generate random bytes securely.
    (SecretKey, test_secret_key, 1, BLAKE2B_KEYSIZE, 32)
}
```

Then we'd convert the key internally to words. That way we avoid forcing users to make a conversion, since secrets keys are usually handled as raw bytes.

Lore (not necessary):

This leads us to why the current BLAKE2b is split across two hazardous modules. Because MACs are by definition also sensitive types, that need constant-time handling and omitted debugs. But let's keep that out of scope for this PR for now, as long as we make the secret key the correct newtype.

If you have any ideas for nice abstraction between those two, I'm all open for suggestions. Feel free to ignore though.

src/hazardous/hash/blake3/mod.rs
```diff
@@ -0,0 +1,157 @@
+// MIT License
+
+// Copyright (c) 2018-2026 The orion Developers
```
brycx: Copyright 2026 only, since this files just got created.

src/hazardous/hash/blake3/mod.rs
```diff
@@ -0,0 +1,157 @@
...
+        self.chunk.update(&data[..take]);
+
+        self.update(&data[take..])
+    }
+
+    fn flush_state(&mut self) {
+        // Get final chaining value
+        let is_root = false;
+        let cv = self.chunk.finalize_chunk(is_root).truncate();
+        self.total_chunks += 1;
```
brycx: I think this needs to be `.checked_add(1).unwrap(1)`. While it's more a theoretical panic/wrap-around, we need it explicit and documented, since it's an algorithmic input-size limit IIUC.

src/hazardous/hash/blake3/mod.rs
```diff
@@ -0,0 +1,157 @@
...
+
+/// Blake3 configuration
+pub struct Blake3 {
+    chunk: ChunkState,
+    chain_values: TreeStack,
+    mode: Mode,
+    total_chunks: u64,
+}
+
+impl Blake3 {
```
brycx: All other streaming-based contexts in Orion have `.reset()` function available. Do you see a sensible way to add this for this case, to keep the API familiar? All fields are modified in-place, so perhaps not?

src/hazardous/hash/blake3/mod.rs
```diff
@@ -0,0 +1,157 @@
...
+// all copies or substantial portions of the Software.
+
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
```
brycx, line 22: Missing documentation, similar to what other hashes/MACs have. You don't need to do this @jakub-mata, but also a reminder to myself.

src/hazardous/hash/blake3/internal.rs
```diff
@@ -0,0 +1,463 @@
+use core::array;
```
brycx, line 23: Missing license.

src/hazardous/hash/blake3/internal.rs
```diff
@@ -0,0 +1,463 @@
...
+    /// Returns the length of the already consumed chunk content in bytes.
+    /// Invariant: len <= CHUNK_LEN
+    pub(crate) fn len(&self) -> usize {
+        BLOCK_LEN * self.blocks_compressed as usize + self.block_len as usize
+    }
+
+    /// Consumes the provided `input`. The size of the input must not be
+    /// greater than CHUNK_LEN - len().
+    pub(crate) fn update(&mut self, input: &[u8]) {
+        assert!(input.len() <= CHUNK_LEN - self.len());
```
brycx: If this is a panic that can occur, then this needs to be a propagated error all the way up to user-reachable API. If not, can it be made `debug_assert!()` so we run it only on tests?

src/hazardous/hash/blake3/mod.rs
```diff
@@ -0,0 +1,157 @@
...
+use crate::hazardous::hash::blake3::cvstack::{FinalizeCommand, PushCommand, TreeStack};
+use crate::hazardous::hash::blake3::internal::{
+    compress, ChunkState, CHUNK_LEN, CHUNK_START, IV, KEYED_HASH, PARENT,
+};
+use core::cmp::min;
+
+/// Blake3 can run in different modes based on usage.
+/// Currently, only Hash and KeyedHash is supported.
+pub enum Mode {
```
brycx: Missing `#[derive(PartialEq, Debug)]`.

src/hazardous/hash/blake3/mod.rs
```diff
@@ -0,0 +1,157 @@
...
+    fn key_words(&self) -> [u32; 8] {
+        match &self {
+            Mode::Hash => IV,
+            Mode::KeyedHash { key } => *key,
+        }
+    }
+}
+
+/// Blake3 configuration
+pub struct Blake3 {
```
brycx, line 101: Missing `#[derive(PartialEq, Debug, Clone)]`.

src/hazardous/hash/blake3/cvstack.rs
```diff
@@ -0,0 +1,158 @@
...
+        self.stack[self.stack_len as usize] = cv;
+        self.stack_len += 1;
+    }
+
+    // Pops the top of the stack
+    //
+    // The popped value is not manually deleted, only the counter is decreased.
+    fn pop(&mut self) -> ChainingValue {
+        self.stack_len -= 1;
+        self.stack[self.stack_len as usize] // Panics if out-of-bounds
```
brycx: Is it reasonable to expect this to panic at any time?

src/hazardous/hash/blake3/mod.rs
```diff
@@ -0,0 +1,157 @@
...
+enum FlagContext {
+    // Indicates START_FLAG plus other flags based on mode
+    Initial,
+    // Indicates PARENT plus other flags based on mode
+    Parent,
+}
+
+impl Mode {
+    fn derive_init_state(&self) -> ChunkState {
+        let key_words = self.key_words();
```
brycx: More general to this PR: there's no zeroization of key-material.

src/hazardous/hash/blake3/mod.rs
```diff
@@ -0,0 +1,157 @@
...
+            self.chain_values.root_output(FinalizeCommand {
+                current_cv: current_state.truncate(),
+                key_words,
+                flags: self.mode.derive_flags(FlagContext::Parent),
+            })
+        };
+
+        reader.fill(out_slice);
+    }
+}
```
brycx, line 215: We want to add a test here at the bottom, that implements `StreamingContextConsistencyTester` found here: `src/test_framework/incremental_interface.rs`.

This is pretty crucial for testing the correctness of incremental handling of inputs.

src/hazardous/hash/blake3/mod.rs
```diff
@@ -0,0 +1,157 @@
...
+
+        if self.chunk.len() == CHUNK_LEN {
+            self.flush_state();
+        }
+
+        let want = CHUNK_LEN - self.chunk.len();
+        let take = min(want, data.len());
+        self.chunk.update(&data[..take]);
+
+        self.update(&data[take..])
```
brycx: This recursive handling here worries me a bit. In release mode, I don't think this would overflow the stack, but in debug it could (not a argument to _not_ remove recursion in crypto-code). If we ever start testing large inputs, this would eventually SIGABRT, depending on the stack for the test thread AFIU.

Have you looked at how we handle this with `while !data.is_empty()`, @jakub-mata? I think that approach would be more robust than this recursion-based one.

brycx (changes requested):
Thank you a lot for all this hard work, @jakub-mata! I get the feeling you were careful and put a lot of effort into it, which really tells.

Let me know if any of my review comments don't make sense or you believe you have a better approach.

src/hazardous/hash/blake3/cvstack.rs
jakub-mata: `stack_len` is updated only in the `push` and `pop` functions so the panics do not occur. I removed the comment.

src/hazardous/hash/blake3/mod.rs
jakub-mata: Done.

src/hazardous/hash/blake3/mod.rs
jakub-mata: I have not written the documentation here yet, but can do so to save you work, so you can just check it.

src/hazardous/hash/blake3/mod.rs
jakub-mata: Done. `Clone` was also added so the `Blake3` struct can have it as well.

src/hazardous/hash/blake3/mod.rs
jakub-mata: Those are the things I know in theory but don't apply in practice :) Hopefully, I've used the abstraction correctly.

src/hazardous/hash/blake3/mod.rs
jakub-mata: I've added zeroization when dropping here, as well as in `TreeStack` and `ChunkState` where the key or other secrets can be derived.

src/hazardous/hash/blake3/mod.rs
jakub-mata: Done.

src/hazardous/hash/blake3/mod.rs
jakub-mata: Actually, it shouldn't be a problem. I simply drop the previous state and assign it the initial one.

src/hazardous/hash/blake3/mod.rs
jakub-mata: That's reasonable, I've removed the recursion here and in the `update()` function inside `ChunkState` (called on line 112 here). A `while` loop is used instead.

src/hazardous/hash/blake3/mod.rs
jakub-mata: Nice catch, done :)

src/hazardous/hash/blake3/mod.rs
jakub-mata: Done, but I've used a tiny workaround. My `finalize()` function consumes the hasher, which I find quite nice as it disallows any further usage. However, the testing trait expects to be *able* to call it, which in my case just cannot happen (I suppose the usage of `mut& self` is expected with an internal flag, something like `is_finalized`). Instead of changing the API, there is a special testing struct which has this flag and an inner hasher to work around it.

I hope that's fine.

tests/hash/blake3_kat.rs
jakub-mata: I see you've found the exact commit for me, I've added it :)

tests/hash/blake3_kat.rs
jakub-mata: Done.

tests/test_data/third_party/blake3_test_vectors.json
jakub-mata: My bad, the official repository should be the source.

src/hazardous/hash/blake3/mod.rs
brycx: That's fine. I'll do it later after this is merged 👍

src/hazardous/hash/blake3/cvstack.rs
```diff
@@ -0,0 +1,185 @@
...
+    CFState, ChainingValue, CompressionFn, OutputReader, BLOCK_LEN, ROOT,
+};
+
+// The maximum possible height of a BLAKE3 tree: the input is capped at 2^64B
+// with 2^10B nodes => 2^54B nodes
+const MAX_TREE_DEPTH: usize = 54;
+
+#[derive(Debug, Clone)]
+pub(crate) struct TreeStack {
+    stack: [ChainingValue; MAX_TREE_DEPTH],
```
brycx, line 33: If the `stack` is zeroized, then it also needs to be an omitted debug.

src/hazardous/hash/blake3/internal.rs
```diff
@@ -0,0 +1,498 @@
...
+
+#[derive(PartialEq, Debug, Clone)]
+pub(crate) struct ChunkState {
+    /// The resulting chaining value of the last compressed block.
+    cv: ChainingValue,
+    /// The number of fully compressed chunks so far.
+    chunk_counter: u64,
+    /// Input buffer for the next block. When full, it is compressed
+    /// and flushed.
+    block: [u8; BLOCK_LEN],
```
brycx, line 76: If secret data flows into here and is zeroized it also needs omitted debug.

src/hazardous/hash/blake3/internal.rs
```diff
@@ -0,0 +1,498 @@
...
+// Permutation done after each round (except for the last one)
+const MSG_PERMUTATION: [usize; 16] = [2, 6, 3, 10, 7, 0, 4, 13, 1, 11, 12, 5, 9, 14, 15, 8];
+
+// 32B intermediate hash output (a.k.a. chaining value)
+pub(crate) type ChainingValue = [u32; 8];
+
+#[derive(PartialEq, Debug, Clone)]
+pub(crate) struct ChunkState {
+    /// The resulting chaining value of the last compressed block.
+    cv: ChainingValue,
```
brycx, line 71: If secret data flows into here and is zeroized it also needs omitted debug.

src/hazardous/hash/blake3/mod.rs
brycx: While you've correctly set up `SecretKey`, it's not being used here. Because `Blake3` derives `Debug`, which `Mode` does as well and `mode` is a field for `Blake3`, a `Debug` call to `Blake3` instance will leak the secret key here.

So this field `Mode.key` either needs to be `SecretKey` instead of array or the parsing/holding of the key needs to be removed entirely from this enum

src/hazardous/hash/blake3/mod.rs
brycx: I completely understand your preference for this type of API, there's less testing even. But it does go against every other streaming API exposed in Orion, which isn't super user-friendly either. I'll have to think about this a bit more.

brycx (changes requested):
I'll have to think about what to do with the `mut self` and `&mut self` a bit more.

The rest is just adding omitted debugs where you've added the zeroization.

src/hazardous/hash/blake3/internal.rs
jakub-mata: Thank you for your patience, I've completely missed that. `cv` can contain the secret key upon initilization so I've omitted the debug.

src/hazardous/hash/blake3/cvstack.rs
jakub-mata: The entire structure is now omitted in debug as it can potentially reduce the complexity of a brute force attack (knowing the intermediate chaining values).

src/hazardous/hash/blake3/internal.rs
jakub-mata: `block` actually just contains the incoming message to be hashed so it's not secret. I've left it in the debug. I've removed the zeroing as well.

jakub-mata:
I'm willing to change the API of `finalize` to accept `&mut self` and change the tests accordingly if the overall API consistency is more important. When you decide, let me know :)

src/hazardous/hash/blake3/internal.rs
brycx: That's perfect, thanks!

src/hazardous/hash/blake3/mod.rs
brycx: This seems to have been taken care of.

brycx:
> I'm willing to change the API of finalize to accept &mut self and change the tests accordingly if the overall API consistency is more important. When you decide, let me know :)

Thank you for sticking with the PR. I'd like to have it use `&mut` instead of the consuming one. Again, I see the benefit of the choice clearly, it just going to stick out like a sore thumb when all others are implemented in another way. 

Apart from this change, I believe this is ready to be merged. I'll add documentation in a follow-up PR.

brycx:
A note for historic design rationales: Because BLAKE3 acts as both a XOF, MAC and Hash, the typical `Tag`-based MAC interface hasn't been applied here. There's no stack-allocatable upper bound on keyed hashes produced and enforcing a `Tag` seems to lock us out of having an API that can be verified in integration tests in `/tests` because of that, since KATs also include large input.

I haven't been able to come up with a reasonable solution for this. Should be considered further for the upcoming `0.18.0` release.

While dangerous and not what you'd expect from other modules in `hazardous`, it is still in that module which also has a warning. The security concern has further been specifically documented in the `Security` docs section.

brycx (approved):
Thank you for all your hard work on this, @jakub-mata! Really awesome, I really appreciate it - thank you.

